PR #26, opened 2015-05-08
Click displayed button element on persona login

theno:
Currently, personas login page considers desktop vs. mobile browsers.  This
results in several login button elements; but only displayed buttons are able
to be clicked by HTML/JavaScript.

(Sometimes the previous code would fail, another day it succeeds.)


hjwp:
Sorry to leave this so long without so much as a response, and thanks very much for taking the time to do a PR.
Persona is now dead (RIP), and the second edition has no mention of it, so am closing this for now.

PR #29, opened 2016-08-24
ch15: explain de-spike

tartley:
A tiny, innocuous suggestion.

My git-diff shows an unexpected '^M' in the change, but I think my line-endings are all kosher. You're using DOS line endings, right? That's what my vim tells me it's saving, and a hex editor says my change has the same line ending as all the other lines, so I think I'm good.


hjwp:
Thanks Jon! Yes, dos line endings are needed by the o'reilly online editor, is annoying.

PR #31, opened 2016-08-26
Tell the reader which html file this is

tartley:
It took me a few moments of grepping back through
the book to figure out which html file this was
that we were editing.

This isn't needed if the filename in the markup surrounding
the source code somehow gets displayed as a caption for the
source code in the rendered book - but I don't see it in the
PDF version I'm looking at.

### DIFF
--- a/chapter_15.asciidoc
+++ b/chapter_15.asciidoc
@@ -130,8 +130,8 @@ Front-end Log in UI
 ^^^^^^^^^^^^^^^^^^^
 
 ((("spiking", "frontend")))
-Let's start with the front-end, just some log in and log out
-links...
+Let's start with the front-end, just some log in and log out links,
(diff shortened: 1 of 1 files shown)


tartley:
Hey Uncle Harry. Reading on, I'm finding this issue (if it is real?) is pervasive in chapter 15. You encourage readers to implement the spike, by creating a branch for it. But then the source code in which you show the spike, you don't specify which source files they are expected to go in. Are you assuming that readers will be sufficiently familiar with Django by this point to figure out sensible places to put the spike changes? Or should each code snippet of the spike have a caption telling the reader which file this change goes into?

hjwp:
HI JB, I just found all these PRs (or maybe I had them flagged for followup ages ago and they just fell off the list).  but in any case, in re: this specific suggestion, the code listings do each have a caption with the filename, so I think we're ok.

PR #33, opened 2016-08-26
Some thoughts about linking to 12-factor

tartley:
I augmented a TODO with some thoughts.


hjwp:
am popping a little hyperlink to 12-factor for now, but if you find anything you prefer do sing out!

PR #34, opened 2016-08-29
Convert REQUIRED_FIELDS from tuple to list

tartley:
A very minor quibble, but in the spirit of lists being semantically like
ordered collections, versus tuples being like structs, it seems to make
more sense to me.

You may have some angle in mind that relies on the immutability,
in which case ignore this.

Best!

### DIFF
--- a/chapter_15.asciidoc
+++ b/chapter_15.asciidoc
@@ -913,7 +913,7 @@ out the answers to these questions for yourself.  Here you go:
 ----
 class User(models.Model):
     email = models.EmailField()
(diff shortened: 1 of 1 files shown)


hjwp:
I like this but is a bit too much hassle for now, cos i'd have to feed thru all the changes via a million cherry picks and stuff.  will put it on my todos list tho.

hjwp:
I only went and did it anyway. #a0618c177e022ded8964a5ac62d6de1f8a17754a

PR #37, opened 2016-09-03
ch12: fix subheading

mbeijen:
This small issue prevented the subheading to render properly, see screenshot below:

schermafbeelding 2016-09-03 om 10 31 13 am

Thanks for the book, it's very nice!


hjwp:
thanks!

PR #40, opened 2016-12-29
Fix FT test result in chapter6

tntC4stl3:
Actuall if one go through the book step by step, it's impossible for him to get an `AssertionError` like this, because we still use the same function `check_for_row_in_list_table` like the first test do.


tntC4stl3:
Sorry, my bad, I lost some code in functional test...
Close this pull request.

hjwp:
Do feel free to send PRs for anything else that you think might need changing though!

PR #41, opened 2016-12-29
Minor edits + typo correction

wasabigeek:


hjwp:
thanks!  I'll keep the symlink relative I think, but I very much appreciate the typo correction.  keep 'em coming!

wasabigeek:
Does the symlink work? I assumed it wouldn't since we weren't in the right
folder

hjwp:
yep!  the symlink target is relative to the symlink itself.  try it:

```bash
cd ~
echo "here" > /tmp/foo
echo "not here" > ./foo
ln -s ./foo /tmp/bar
cat /tmp/bar # should show "here"
```

PR #42, opened 2017-01-30
This was removed in 201670dc3e, presumably by accident

tartley:


tartley:
This PR restores the end of a sentence that was deleted in a recent commit. I assume the deletion was an error.

hjwp:
gotcha.  was kinda hard to see inamongst all that nested footnote thingie.  thanks JB!

PR #43, opened 2017-01-30
minor rewording suggestion

tartley:
I think it's helpfully more specific to say that 'our test' is doing
the mocking, rather than 'we're asking Python to' do it. YMMV


tartley:
Dammit, plus I see some diffs to ascidoctor.css got committed. I didn't make those deliberately - they must happen as part of building the html output. I'll revert that. Sorree!

hjwp:
Grand.  thanks Jonathan!

PR #47, opened 2017-02-07
Allow less secure apps

tartley:
As I understand it: The book's current advice to create app-specific passwords in your Google account is only correct if your Google account uses 2-factor authentication.

This PR advises readers of what to do if they don't use 2-factor. In this case, creating app-specific passwords is not possible, and SMTP requests are refused unless you configure your google account to "allow access for less secure apps"


hjwp:
left this too long to merge automatically but am going to do so now, somebody else was complaining about less secure apps just the other day... thanks again jonno.

PR #48, opened 2017-02-07
Fix markup of __init__.py

tartley:
This was rendered in HTML as

    _\_init__.py

This commit makes it render as

    __init__.py (in italics)

which I think is desired.


hjwp:
annoyingly oreilly themselves use a slightly different toolchain.  i think i've made it into a cross-compatible (but hella ugly) version.

PR #51, opened 2017-03-20
Updated output of *pip install "django==1.11b1"...

jwm-evans:
Changed from Django-1.11.5-py2.py3-none-any.whl to Django-1.11b1-py2.py3-none-any.whl


hjwp:
Great!  thanks for this too :)

PR #52, opened 2017-03-21
Fixed typo in systemctl

jwm-evans:
systemctl was misspelled  as systemctrl after enabling debugging in gunicorn


hjwp:
Thanks very much!

PR #53, opened 2017-03-24
Change Django version in requirements.txt

sfarzy:
Django 1.11.5 is invalid, side effect of bulk updating. :wink: 
Changed it into 1.11.b1


hjwp:
thanks!

PR #54, opened 2017-03-24
Corrected number of lines replaced with assertIn

jwm-evans:
In the older text (which didn't make use of the new "f-string" syntax), the assertTrue(...) for the first list item took up six lines. The new notation reduced this to four, and using the assertIn(...) test further reduced it to one.


hjwp:
Nice attention to detail, thanks James!

PR #55, opened 2017-03-24
Grammatical fix, changing 'so' to 'but'

jwm-evans:
The first two parts  of the sentence suggest the most probable action to be taken, whereas the last part suggests an alternative. In this situation I feel that 'but' would be a more appropriate conjunction


hjwp:
I guess what I'm trying to say is, "so if in doubt, try the unit tests first and then the functional tests".  but that's a bit long-winded.  will have a think about how to rephrase...

PR #58, opened 2017-03-29
Corrected minor typo -'a' should be 'and'

jwm-evans:


hjwp:
can't tell you how much i appreciate this.  thanks James!

jwm-evans:
I have a bunch more that I marked up in the PDF. I'll probably start
submitting those in the evening.

PR #67, opened 2017-04-13
Use from django.urls import resolve

hunterji:
As of Django 1.10.0 django.core.urlresolvers.resolve has been replaced
with django.urls.resolve.

See https://docs.djangoproject.com/en/1.11/ref/urlresolvers/


hjwp:
Great!  Thanks, I didn't know about this, much nicer.  It's going to take a bit more work to apply this throughout the book, but this is a good start :smile:

hjwp:
actually not all that much work!  all done now, thanks again.

PR #68, opened 2017-04-13
update chapter number scopes of book parts

das-g:
update chapter number scopes of book parts to match the current state as of bab0b9b6993a14f82efecb38f7bc4804f986b2c3

(I've also counted the [Epilogue](http://www.obeythetestinggoat.com/book/epilogue.html), as that is a numbered chapter in the TOC. I didn't count the appendices although the TOC makes it seem as if they blong to part 3, too.)

You might want to check whether the content descriptions of the parts still matches, too. (I've just started reading the book, so I can't know.)

### DIFF
--- a/preface.asciidoc
+++ b/preface.asciidoc
@@ -162,19 +162,19 @@ Outline
 
 I've split this book into three parts.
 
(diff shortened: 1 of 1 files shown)


hjwp:
this is great thanks!

PR #69, opened 2017-04-21
Typo in A Simple Form chapter

hunterji:
...Django has add the... -> ...Django has added the...


hjwp:
Brilliant, thanks very much! :)

PR #73, opened 2017-06-23
Update chapter_advanced_forms.asciidoc

gonulate:
Added the class names for the new functions.

It might be helpful to see the class names. If your reason for not to insert them is pedagogical, it is understandable.


hjwp:
thanks for the suggestion!  i'm too close to sending the book to the printers to merge this now, but will definitely put it on my list for the next minor release :+1:

PR #78, opened 2017-09-11
Format code blocks in doc

salmanulfarzy:


hjwp:
amazing!  are you actually planning on running the book's tests?  I don't think anyone apart from me has ever done that.   Maybe one of my tech reviewers tried once...

PR #81, opened 2017-10-22
Fixes typo

wayneadams:
Changes one of the example subdomains to actually be superlists-staging.ottg.eu


hjwp:
haha how on earth did i manage to write that?  thanks very much!

PR #87, opened 2018-02-14
Add missing line of code in example of Chapter 25

wwarne:
In chapter 25 we start using the Page object to rewrite our test. This line was in the first code example and wasn't in the second. It's a little confusing because it seems like we can delete it but it will be a mistake.


hjwp:
Yes!  thanks very much :)

hjwp:
how did you like the book btw?  drop me an email to obeythetestinggoat@gmail.com... 🤙

PR #88, opened 2018-02-20
Minor typo causes unit test to fail

jwm-evans:
The test "test_page_returns_correct_html" (prior to being refactored) expects the title element to be "To-Do Lists", but the template file has "lists" as all lower case. Correcting the home.html file allows the test to pass

### DIFF
--- a/chapter_philosophy_and_refactoring.asciidoc
+++ b/chapter_philosophy_and_refactoring.asciidoc
@@ -359,7 +359,7 @@ https://docs.djangoproject.com/en/1.11/intro/tutorial03/#write-views-that-actual
 [source,html]
 ----
 <html>
-    <title>To-Do lists</title>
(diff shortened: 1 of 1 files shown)


hjwp:
I think I use a lowercase "l" pretty consistently in all the code examples, both tests and html.  That applies to `test_home_page_returns_correct_html`, both when I introduce it in the previous chapter, and when we refactor it in this chapter.

Could you have made a mistake/typo when you first created that test?

jwm-evans:
I think I did. I only realised that after submitting the pull request.
Sorry for the false request

hjwp:
no problem!  feel free to close the pull request then, and please don't hesitate to submit another one if you spot anything else -- even if you're not quite sure, i'd rather have too many than too few :)

PR #89, opened 2018-02-27
Missing word in the overview

jwm-evans:
There is a word missing in the overview for the chapter on "moving to a production-ready config", specifically between "to" and "DEBUG"  in the Security bullet-point. "set" seems like a good fit.


hjwp:
just found this -- thanks!

PR #90, opened 2018-03-08
Update chapter_mocking.asciidoc

jwm-evans:
The original phrasing (..getting the boilerplate for a basic URL and view in, via...) makes it look like a word is missing.


hjwp:
thanks!  i think it's clearer now :)

PR #91, opened 2018-03-15
Unnecessary 'as'

jwm-evans:
The word 'as' does not belong in the sentence "patch decorators are applied innermost first, so the new mock is injected to our method **as** before the mockNewListForm"


hjwp:
briliant!  thanks :)

PR #96, opened 2018-06-29
Added link to learn more about regular expression

vaibhavhrt:
Since it's the (I think) first time book introduces regular expressions, there should be a link to learn more about regular expressions.


hjwp:
Thanks!  I feel like most anyone reading this book will probably know what regular expressions are though? and they can always google themselves if needed...

PR #97, opened 2018-08-06
typo + missing a "]" in a footnote

patjouk:
The footnote in chapter 10 is not rendered correctly ;)

> Hooray, a change that went without a hitch for once!footnote:[If you’re using Fedora/CentOS, you may run into an issue with private tmmp directories. more info here Moving on…​

Thanks again for the great book!


hjwp:
Yes!  thank you :)

PR #98, opened 2018-08-10
Fixed a typo

DarkSuniuM:
Well, It's something :|


hjwp:
hi there, I think that `<3>` is correct actually.  callout number two is for the test method, and number three appears twice, once for setUp and once for tearDown

[screenshot from book](https://postimg.cc/image/b71mfxl47/)

hjwp:
please don't hesitate to send any other suggestions!  I'd much rather have too many and say no now and again, than too few...

PR #106, opened 2019-04-03
Update chapter_02_unittest.asciidoc

MarckK:
`ResourceWarning` no longer emitted.


hjwp:
great! always nice to make a change that simplifies the book.  Thanks very much!

PR #109, opened 2019-05-14
Redundant "it"

jwm-evans:
There is a redundant "it" on line 631


hjwp:
belated thanks!  :smile:

PR #112, opened 2019-10-04
added links to all the numbered chapters

dlim87:
Added a link to the next chapter near the end of each of the numbered chapters


hjwp:
this is a brilliant idea but i can't merge it as-is because these changes would also go into the printed book, where it doesn't make sense to add cross-refs to later chapters.

there may be a way to hack in a link to the next chapter in the script that generates the book html though? https://github.com/hjwp/Book-TDD-Web-Dev-Python/blob/master/copy_html_to_site_and_print_toc.py

PR #121, opened 2019-12-24
Item numbers got flipped in the test code

tsvenson:
Gave me some headache, trying to figure out if they where flipped on purpose or not. But, since i couldn't find any clues further on in the chapter I think this is a typo.

### DIFF
--- a/chapter_explicit_waits_1.asciidoc
+++ b/chapter_explicit_waits_1.asciidoc
@@ -462,8 +462,8 @@ Now we can rename our method calls, and remove the voodoo ++time.sleep++s:
     inputbox.send_keys(Keys.ENTER)
 
     # The page updates again, and now shows both items on her list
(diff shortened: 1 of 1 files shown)


hjwp:
just saw this. did you try it in your own code?  we're checking for the (new) second row first, and then we check that the old (first) row is still there.

tsvenson:
Oh, I see now what got me confused. On page 91 the order of the two got switched compared to page 65.

Page 65:
`# The page updates again, and now shows both items on her list
self.check_for_row_in_list_table('1: Buy peacock feathers')
self.check_for_row_in_list_table('2: Use peacock feathers to make a fly')`

From page 91:
`# The page updates again, and now shows both items on her list
self.wait_for_row_in_list_table('2: Use peacock feathers to make a fly')
self.wait_for_row_in_list_table('1: Buy peacock feathers')`

I still have the order from Page 65. Does that matter?

PR #123, opened 2020-05-27
user can disregard droplet options

russHyde:
Digital ocean droplets have options for configuring block storage, a VPC network, IPv6, user-data, monitoring and back-ups when creating a droplet. These can be ignored (hopefully! I've just ignored them and I've never setup a web server before)


hjwp:
brilliant, thanks!

PR #144, opened 2023-05-19
review pre-requisite-installations

Xronophobe:
- change single quoted 'and' -s to italic
- add notes on installing pyenv
- update virtualenv directory name to .venv
- add notes on different pip install output
- update ImportError message
- update tdd-python-book to goat-book

### DIFF
--- a/pre-requisite-installations.asciidoc
+++ b/pre-requisite-installations.asciidoc
@@ -134,7 +134,7 @@ The Git version control system::
     ((("Git", "downloading")))
 
 
-A virtualenv with Python 3.11, Django 4.2, and Selenium 3 in it::
+A virtualenv with Python 3.11, Django 4.2, and Selenium 4 in it::
     Python's virtualenv and pip tools now come bundled with Python (they
     didn't always used to, so this is a big hooray).  Detailed instructions for
     preparing your virtualenv follow.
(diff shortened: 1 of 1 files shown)


hjwp:
lovely thanks for these Csanad!

hjwp:
any general comments on the chapter?  I think @ritaf-ORM was hoping for some high-level / overall reactions for each one?

Xronophobe:
Right. I took it as **if** I have comments/concerns/feedback on the chapter, then I should provide them in the PR (and also email it to Rita). I had some thoughts here and there, which I left in the source code, as comments.
I will, however elaborate briefly and summarize my thoughts on the chapter later, in the PR description. Is it fine if (beyond the email) I would just update this already closed and merged PR's description?

hjwp:
thanks buddy!  i would say - don't duplicate effort.  if you've made comments inline, that's fine, no need to re-summarise them really.

just if you have a sort of overall reaction to the chapter - even just a single sentence, then put it in the description.  

and also, sometimes if you have a more general comment on the chapter or on the book so far, something that's not really a comment on a specific line of text or code, then the pr description is a good place for it.

Xronophobe:
@hjwp the links are at the bottom, look for the "Project supported by" text

hjwp:
ah yes.  definitely sketchy.  will switch to the https://diveintopython3.problemsolving.io/ url at least.

hjwp:
but yeah we probably need a new and better free intro to python 3...

Xronophobe:
Even though I would prefer suggesting a book, eg. Codecademy offers a seemingly good free beginner course. Unfortunately, they are not covered on tosdr.org or tldrlegal.com and I haven't had the time yet to skim through their privacy policy (just before suggesting an online service that requires users to sign up).

hjwp:
am gonna link to https://www.manning.com/books/the-quick-python-book-third-edition instead.

PR #145, opened 2023-05-20
review chapter_01

Xronophobe:
This chapter is easy to follow. I like the explanation on refraining of using easy-to-test but difficult-to-apply-IRL examples and instead, building an actual website.

Maybe I would move the 2nd footnote higher up, or even mention it in the pre-requisites chapter. There is a brief description for setting up git for Windows, maybe it would fit there. As I was reviewing this chapter, I actually started writing a suggestion about providing some help with setting up a newly installed git, when I noticed there already was a footnote about it. Maybe it isn't very obvious to spot first.


chapter_01.asciidoc
```diff
@@ -73,9 +73,11 @@ from selenium import webdriver
 browser = webdriver.Firefox()
 browser.get("http://localhost:8000")
 
-assert "Django" in browser.title
+assert "Congratulations!" in browser.title
 ----
 ====
+// CSANAD: They changed the default title for Django, it currently says
+//      "The install worked successfully. Congratulations!"
```
hjwp, line 80: oh i'm going to have to take a look at that!  there's supposed to be a test that checks for this...

hjwp:
hmmm  why isn't CI running on this PR i wonder?

chapter_01.asciidoc
hjwp, line 80:
```suggestion
assert "Django" in browser.title
----
====

// CSANAD: They changed the default title for Django, it currently says
//      "The install worked successfully. Congratulations!"
// TODO - harry to amend code listing and adjust tests.
```

PR #149, opened 2023-05-22
review  chapter_unit_test_first_view

Xronophobe:
Overall, this was a pretty long chapter, but I don't think it would make sense to split into maybe two smaller ones.

### DIFF
--- a/chapter_unit_test_first_view.asciidoc
+++ b/chapter_unit_test_first_view.asciidoc
@@ -3,7 +3,7 @@
 
 
 We finished the last chapter with a functional test failing,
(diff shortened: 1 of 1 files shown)


chapter_unit_test_first_view.asciidoc
```diff
@@ -3,7 +3,7 @@
 
 
 We finished the last chapter with a functional test failing,
-telling us that it wanted the home page for our site to have ``To-Do'' in its title.
+telling us that it wanted the home page for our site to have ``Congratulations!'' in its title.
```
hjwp, line 6:
```suggestion
 telling us that it wanted the home page for our site to have ``To-Do'' in its title.
```

PR #150, opened 2023-05-23
Comments on Chapter 2

seddonym:
Everything worked for me.

Really enjoying it so far, I like the feeling that we're pair-programming.

### DIFF
--- a/chapter_02_unittest.asciidoc
+++ b/chapter_02_unittest.asciidoc
@@ -231,8 +231,6 @@ class NewVisitorTest(unittest.TestCase):  # <1>
         self.browser.quit()
 
     def test_can_start_a_todo_list(self):  # <2>
-        // DAVID: Pedantry: should we be consistent with whether
-        // comments end with a full stop?
         # Edith has heard about a cool new online to-do app.
         # She goes to check out its homepage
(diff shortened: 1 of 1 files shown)


chapter_02_unittest.asciidoc
```diff
@@ -229,6 +231,8 @@ class NewVisitorTest(unittest.TestCase):  # <1>
         self.browser.quit()
 
     def test_can_start_a_todo_list(self):  # <2>
+        // DAVID: Pedantry: should we be consistent with whether
+        // comments end with a full stop?
```
hjwp, line 235: reminder to avoid suggesting changes directly into code unless absolutely necessary.  will re add below.
```suggestion
```

chapter_02_unittest.asciidoc
```diff
@@ -254,7 +258,7 @@ You'll probably notice a few things here:
 <1> Tests are organised into classes, which inherit from `unittest.TestCase`.
```
hjwp, line 258:
```suggestion
// DAVID: Pedantry: should we be consistent with whether
// comments end with a full stop?

<1> Tests are organised into classes, which inherit from `unittest.TestCase`.
```

PR #151, opened 2023-05-23
Review pre-requisite-installations.asciidoc

jangia:
Add 2 suggestions:
- About Note on IDEs
- About locking Selenuim version

### DIFF
--- a/pre-requisite-installations.asciidoc
+++ b/pre-requisite-installations.asciidoc
@@ -92,6 +92,9 @@ There's a little bit of JavaScript in the second half of the book.  If you
 don't know JavaScript, don't worry about it until then, and if you find
 yourself a little confused, I'll recommend a couple of guides at that point.
 
+// JAN: This note seems a bit unnecessary. Nowdays I really don't know of anyone that wouldn't use IDE. They're too useful to not use them.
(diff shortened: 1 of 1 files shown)


pre-requisite-installations.asciidoc
```diff
@@ -365,7 +368,9 @@ activate script that was compatible with Git-Bash. Reinstall the latest Python
 
 *******************************************************************************
 
-
+// JAN: I'd suggest to use exact Selenium version as well for two reasons
+// 1. To avoid readers issues because of breaking changes in Selenium updates. I had a problem with poetry where installing the latest versions instead of specific one broke the whole course.
+// 2. To not raise eyebrows - why is there version for Django but not for selenium
```
hjwp, line 373: thanks for this too! 

So the problem with pinning selenium, in my experience from previous editions, is that selenium is in a race to keep up with browsers.  So newer versions of firefox sometimes need a newer selenium, and we cant ask readers to pin firefox
So instead what i do is try and always use the latest -- In CI, the book tests install selenium fresh every time, latest version.  (and django too actually but that's a separate issue)

but you're right, the discrepancy needs explaining.  i'll add a note.

PR #152, opened 2023-05-23
Update chapter_01.asciidoc

jangia:
- Update `master` -> `main`
- Added comment about "Why TDD?"
- Added comment about "Red-Green-Refactor"
- Added comment about failing test due to "Django" string not being present inside default page title

### DIFF
--- a/chapter_01.asciidoc
+++ b/chapter_01.asciidoc
@@ -6,6 +6,8 @@ It's a discipline, like a martial art, and just like in a Kung Fu movie,
 you need a bad-tempered and unreasonable master to force you to learn the discipline.
 Ours is the Testing Goat.
 
+// JAN: As a reader, I have no idea why would I practice TDD in the first place. Why would I want to master this discipline? I think it would be worth mentioning here. Doesn't have to be very long.
+// Maybe something like: "TDD forces you to develop in small chunks. Therefore, you can focus on one thing at the time. As a byproduct, you have executable specification of your features in form of tests. Want to make sure your changes haven't broken any existing feature? Just run the tests. Fast and easy. And so is releasing and deployment when you really practice TDD - when you obey the testing goat"
 
 === Obey the Testing Goat! Do Nothing Until You Have a Test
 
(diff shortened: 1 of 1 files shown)


chapter_01.asciidoc
```diff
@@ -330,7 +336,7 @@ Next we can add the rest of the contents of the current folder, ".":
 ----
 $ pass:quotes[*git add .*]
 $ pass:quotes[*git status*]
-On branch master
+On branch main
```
Xronophobe, line 339: The default branch is still called `master` - at least on my fresh Linux Mint install in a VM in which I'm testing the book's instructions

chapter_01.asciidoc
```diff
@@ -6,6 +6,8 @@ It's a discipline, like a martial art, and just like in a Kung Fu movie,
 you need a bad-tempered and unreasonable master to force you to learn the discipline.
 Ours is the Testing Goat.
 
+// JAN: As a reader, I have no idea why would I practice TDD in the first place. Why would I want to master this discipline? I think it would be worth mentioning here. Doesn't have to be very long.
+// Maybe something like: "TDD forces you to develop in small chunks. Therefore, you can focus on one thing at the time. As a byproduct, you have executable specification of your features in form of tests. Want to make sure your changes haven't broken any existing feature? Just run the tests. Fast and easy. And so is releasing and deployment when you really practice TDD - when you obey the testing goat"
```
hjwp, line 10: I guess I tried to cover some of the "why tdd" stuff in the [preface](https://github.com/hjwp/Book-TDD-Web-Dev-Python/blob/main/preface.asciidoc)

jangia: I see. In that case, just ignore the comment 👍

chapter_01.asciidoc
jangia: On my MacOS, it's main.

Xronophobe: Weird, on my Mac (v13.3.1) it's still `master`. I have git v2.40.0 which is just a minor behind the current latest. Maybe it's customized on yours? What do you get when you enter `git config --global init.defaultBranch` ?

jangia: It says main, but I don't remember changing that. I really don't customize such things. Anyhow, if it's master for others, then let's keep master.

PR #155, opened 2023-05-31
Update link to 2nd edition of Pytest book

jwm-evans:
Brian Okken wrote second edition of "Python Testing with pytest". I have updated the link to the new book


hjwp:
thanks!

PR #158, opened 2023-06-04
review Post and Database

Xronophobe:
It was a pretty long chapter, but I have only found some minor things I think should be changed or rephrased.


hjwp:
brilliant thanks for this Csanad great comments.

chapter_post_and_database.asciidoc
```diff
@@ -190,6 +190,8 @@ image::images/twp2_0501.png["Django DEBUG page showing CSRF error"]
 If you've never heard of a _Cross-Site Request Forgery_ exploit, why not look it up now?
 Like all security exploits, it's entertaining to read about,
 being an ingenious use of a system in unexpected ways.
+// CSANAD:  I think it would be a good idea to suggest using OWASP as a reference for IT
+//          security: https://owasp.org/www-community/attacks/csrf
```
hjwp, line 194: it's already the number 1 result on google...

chapter_post_and_database.asciidoc
```diff
@@ -1248,6 +1266,11 @@ def home_page(request):
     )
 ----
 ====
+// CSANAD:  on the 4th line:
+//          `item.text = request.POST["item_text"]`
+//          we accessed item_text using the get() method previously:
+//          `item.text = request.POST.get("item_text", "")`
```
hjwp, line 1272: yes it's a bit janky.  you can see the request.POST.get is still in the codeblock lower down.

chapter_post_and_database.asciidoc
```diff
@@ -1690,6 +1718,7 @@ Ran 1 test in 5.036s
 
 OK
 ----
+// CSANAD:  The test will actually fail because of `self.fail("Finish the test!")`
```
hjwp, line 1721: hmmm i'll need to check this.

chapter_post_and_database.asciidoc
```diff
@@ -1765,6 +1794,7 @@ Where are we?  How is progress on our app, and what have we learned?
 
 * And we've used at least three different FT debugging techniques: in-line
   print statements, ++time.sleep++s, and improving the error messages.
+// CSANAD:  I can't find where we used in-line print, maybe it's been removed?
```
hjwp, line 1797: good point!

hjwp, line 1797:
```suggestion
* And we've used two different FT debugging techniques:
  ++time.sleep++s, and improving the error messages.
```

chapter_post_and_database.asciidoc
hjwp, line 194:
```suggestion
```

hjwp, line 194:
```suggestion
```

chapter_post_and_database.asciidoc
Xronophobe: I'll double check later tonight but I can't recall commenting that line out in this version of the book yet.

Xronophobe: Ok, so what I found is this `self.fail` is still there in chapter_philosophy_and_refactoring
https://github.com/hjwp/book-example/blob/3d1c41518597cc43f18c1d8a8ac86967998ebc3b/functional_tests.py

But it's been removed from chapter_post_and_database
https://github.com/hjwp/book-example/blob/335a69f69ec9dfac6fe567dcab47cbfbcbc27968/functional_tests.py

I couldn't find this change in either of the chapters' text though.

chapter_post_and_database.asciidoc
```diff
@@ -365,7 +378,7 @@ The notation is `{{ ... }}`, which displays the object as a string:
     {% csrf_token %}
   </form>
   <table id="id_list_table">
-    <tr><td>{{ new_item_text }}</td></tr>  <1>
+    <tr><td>{{ new_item_text }}</td></tr>  # <1>
```
hjwp, line 381:
```suggestion
    <tr><td>{{ new_item_text }}</td></tr>  <1>
```

hashes don't work as comments in html.  it renders ok in the book anyways.

chapter_post_and_database.asciidoc
hjwp: ok the change is in the listing here at ch05l015.  i'll change the text to make it clearer that we are deleting the self.fail.  thanks buddy.

PR #159, opened 2023-06-05
Review chapter_philosophy_and_refactoring.asciidoc

jangia:
- Add note about FT abbreviation
- Add note about assertTemplateUsed test

### DIFF
--- a/chapter_philosophy_and_refactoring.asciidoc
+++ b/chapter_philosophy_and_refactoring.asciidoc
@@ -156,7 +156,6 @@ NOTE: One of the great things about TDD is that you never have to worry about
 
 ``Finish the test'', it says, so let's do just that!  Open up
 'functional_tests.py' and we'll extend our FT:
-// JAN: I suggest to not use abreviation. For you and me, it's obvious what does FT mean but for a reader it might be confusing. FT -> free throw?
 
 [role="sourcecode"]
 .functional_tests.py (ch04l001)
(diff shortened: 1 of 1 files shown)


chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -593,6 +594,7 @@ class HomePageTest(TestCase):
...
... - makes complete sense. But I don't think using self.assertTemplateUsed(response, "home.html") makes it much better. We go from HTML implementation details (self.assertContains(response, "<title>To-Do lists</title>")) to Django implementation details (self.assertTemplateUsed(response, "home.html")). This test will pass whenever I'm using "home.html" template. It may be empty or complete garbage - this test will pass as long as the naming is correct. On the other hand, if I change the name of template, I'll have to update the test. Logic and ouput might stay completely the same. (I wouldn't be surprised if something like this would happen also when changing template engine). In other words, this test is not resistant to refactoring while also not protecting much against regression. Therefore, I'd consider it as low quality test - it won't protect me much while posibly failing when things still work exactly as expected. I know that there's not much breathing room there to actually change much. Nevertheless, I think it's worth bringing this up. (I derived this approach for test quality evaluation from this book: https://www.manning.com/books/unit-testing?query=%20Vladimir%20Khorikov)
```
hjwp, line 597: hmmm good point.

what do you tend to do for unit tests of GET endpoints, that just render templates?

hjwp: OK fair enough.  Being tightly coupled to Django is one of the main drawbacks of the book really, and it's hard to avoid.  The second book is the antidote!

PR #161, opened 2023-06-07
Review philosophy chapter

seddonym:
Really enjoyed the discussion here.

As mentioned in one of my comments, I wonder if it's worth digging a bit more into where the division of unit vs functional tests is here. In practice I would say the functional test is good enough for testing the url routing.

### DIFF
--- a/chapter_philosophy_and_refactoring.asciidoc
+++ b/chapter_philosophy_and_refactoring.asciidoc
@@ -94,7 +94,7 @@ when the simple function gets a tiny bit more complex--perhaps
 it grows an `if`.
 Then a few weeks later it grows a `for` loop.
 Before you know it, it's a recursive metaclass-based polymorphic tree parser factory.
-But because it's had tests from the very beginning,
+But because its had tests from the very beginning,
 adding a new test each time has felt quite natural,
 and it's well tested.
 The alternative involves trying to decide when a function becomes ``complicated enough'',
@@ -104,6 +104,7 @@ it feels like that much more effort to start,
 and you're tempted each time to put it off a little longer,
 and pretty soon--frog soup!
 
+// DAVID: Worth breaking up that last sentence? It's a bit difficult to read.
 
 Instead of trying to figure out some hand-wavy subjective rules
 for when you should write tests,
@@ -114,6 +115,9 @@ you have to take the time to learn the rules before you can break them.
 **********************************************************************
 
 Now, let us return to our muttons.
+
+// DAVID: FYI I had to look up what 'return to our muttons' means.
+
 ((("", startref="TDDneed04")))
 
 [role="pagebreak-before less_space"]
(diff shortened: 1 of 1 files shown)


chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -94,7 +94,7 @@ when the simple function gets a tiny bit more complex--perhaps
 it grows an `if`.
 Then a few weeks later it grows a `for` loop.
 Before you know it, it's a recursive metaclass-based polymorphic tree parser factory.
-But because it's had tests from the very beginning,
+But because its had tests from the very beginning,
```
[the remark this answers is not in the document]
seddonym: Say, you're right - thanks.

I agree your version reads slightly better.

hjwp, line 97:
```suggestion
But because it's had tests from the very beginning,
```

chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -114,6 +115,9 @@ you have to take the time to learn the rules before you can break them.
 **********************************************************************
 
 Now, let us return to our muttons.
+
+// DAVID: FYI I had to look up what 'return to our muttons' means.
```
hjwp, line 119: yes it's very french.

chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -214,6 +218,8 @@ class NewVisitorTest(unittest.TestCase):
 ----
 ====
 
+// DAVID: I'd use ... over [...] in code samples as it's simpler (and still valid Python).
```
hjwp, line 221:
```python
[...]
```

is valid python too!

not sure if i have the energy to go change these and *all* their test. but otoh, i've got so used to the existing ones that i can't "see" them impartially any more.  maybe this is exactly the reason to have fresh eyes.  arg but so much effort.

chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -277,6 +283,9 @@ selenium.common.exceptions.NoSuchElementException: Message: Unable to locate
 element: h1
 ----
 
+// DAVID: Could we take artistic license and break this line after NoSuchElementException:?
+// It would read more nicely.
```
hjwp, line 287: no, the tests are based on running with an 80-character terminal width, so i have to break things exactly that way, pretty much...  exceptions are possible!  but they would be equivalent to putting a hashtag-ignore on that line, ie making it untested...

chapter_philosophy_and_refactoring.asciidoc
```diff
@@ -489,6 +504,11 @@ INSTALLED_APPS = [
 ----
 ====
 
+ // DAVID: This could just be "lists", which is more consistent with the other apps.
+ // See https://docs.djangoproject.com/en/4.2/ref/applications/#configuring-applications
+ //
+ // > by default, if Django finds exactly one AppConfig subclass in the apps.py submodule,
+ // > it uses that configuration for the application.
```
hjwp, line 511: yes i would prefer it.  i actually changed it because the new django tutorial told me to, iirc.

chapter_philosophy_and_refactoring.asciidoc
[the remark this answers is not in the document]
hjwp: thanks both!

PR #167, opened 2023-06-18
Feedback on prerequisites chapter & small typo fix

Enforcer:
# General feedback

Hey 👋 Nice intro to many basic tools necessary for software development with Python.

## Choosing Django
I'm well aware this book is mostly an update & extension of previous versions, but I wonder if focusing solely on Django is the best idea.

Microframeworks are now on the rise (at least in my information buble). I think it could be valuable to also somehow mention how to apply pieces of advice here and there to other frameworks.


hjwp:
thanks very much Sebastian!

I tend to agree with you about the Django focus -- like anything it has pros and cons, it's a tradeoff. So on the plus side this gives people a decent intro to what is still probably the most popular Python web app framework.  but on the downside it does mean a lot of the content is django-specific, and thus less about TDD per se.

but it's just too much effort to change at this stage.  it would be a different book.

thnaks for the fixes, and let me know how you enjoy the rest of the book!

PR #170, opened 2023-06-18
Review of chapter 03

Enforcer:
# General feedback
Informative intro to unit tests of a single view function, can't wait for the next chapter. 

Left a few comments in the body, mostly about the completeness of code snippets and one about comparing unit vs functional tests.

### DIFF
--- a/chapter_unit_test_first_view.asciidoc
+++ b/chapter_unit_test_first_view.asciidoc
@@ -164,6 +164,7 @@ Enough theory for now—let's see how it looks in practice.
 |===
 
 // DAVID: Could include something about how unit tests provide more 'signal' about what went wrong.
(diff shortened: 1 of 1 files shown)


chapter_unit_test_first_view.asciidoc
```diff
@@ -727,6 +729,8 @@ class HomePageTest(TestCase):
 // reading experience if the test method we've already seen was replaced by a `...`. Though
 // it probably would need an explanation that it's a convention we're using.
 
+// SEBASTIAN: Also, I'd suggest adding an import at the top of the snippet `from django.test import TestCase`
```
hjwp, line 732: thanks.  django puts that import into tests.py for you by default, i think that's why i didn't bother including it...

PR #172, opened 2023-06-19
Review of chapter chapter_post_and_database.asciidoc

Enforcer:
Another round of review from me.

Out of curiosity (I'm not a native English speaker) - what's with using "her" pronoun when gender is unknown? I only know about using they/them in such a situation.
I'm talking about a sentence that ends the first paragraph:
> so that we can save it somehow and display it back to her later

# General feedback
This chapter is quite long and reviewing it was a bit exhausting. Is there an option perhaps to split it somehow? Or put some summary in the middle to let a reader catch a breath?

### DIFF
--- a/chapter_post_and_database.asciidoc
+++ b/chapter_post_and_database.asciidoc
@@ -2,7 +2,7 @@
 == Saving User Input: Testing the Database
 
 ((("user interactions", "testing database input", id="UIdatabase05")))
-So far we've managed to return a static HTML page with an input box in it,
+So far we've managed to return a static HTML page with an input box in it.
 Next we want to take the text that the user types into that input box and send it to the server,
 so that we can save it somehow and display it back to her later.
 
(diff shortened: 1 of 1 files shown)


hjwp:
hey!  some people like to alternate between "he" and "her" pronouns when describing generic-example-people, because "they" feels awkward and impersonal, and the default we all tend to slip into otherwise (especially in a male-dominated world like programming) is to just use "he" everywhere...  so i try and keep the ratio about 50/50.

chapter_post_and_database.asciidoc
```diff
@@ -1889,4 +1908,7 @@ The scratchpad to-do list::
 place to write down things that occur to us as we're coding, so that
     we can finish up what we're doing and come back to them later.
 
+// SEBASTIAN: (idea) alternative to maintaining a scratchpad could be to write empty unit tests without implementation.
+//    Such "tests prototypes" could be skipped initially until we work on them.
```
hjwp, line 1912: i've always like this idea in theory, but i've never found myself actually bothering to do it in practice, or not more than once or twice -- have you ever got into the habit?

Enforcer:
> hey! some people like to alternate between "he" and "her" pronouns when describing generic-example-people, because "they" feels awkward and impersonal, and the default we all tend to slip into otherwise (especially in a male-dominated world like programming) is to just use "he" everywhere... so i try and keep the ratio about 50/50.

Oh, now I understand - thanks for taking the time to respond to that!

ritaf-ORM:
The use of pronouns in examples have shifted over the past few years. Nowadays, it's part of our house style to use the generic pronouns they/their when needed, not he, she, or he/she.

PR #175, opened 2023-06-29
Working thru TR suggestions

hjwp:


chapter_post_and_database.asciidoc
```diff
@@ -746,7 +725,6 @@ other methods for your own purposes:
 ----
 ====
 
-// SEBASTIAN: Why not use underscore to emphasize these methods are helper & pretty much private at the same time?
```
hjwp, line 749: @Enforcer as "luck" would have it i'm using unittest at work agin these days, and so I found myself considering this exact question, since i'm much more into using underscorey methods to identify private things these days.

in the end i decided it's not really necessary in test classes, because by convention, we already know that the "public" methods are the test methods that start with `test_`, so anything else is already "private".

i mean, unless you start using inheritance in your test classes,  but then you're in deep deep trouble anyways.

so in this case i decided the underscores aren't worth the readability hit...

chapter_post_and_database.asciidoc
```diff
@@ -1213,41 +1176,106 @@ def test_can_save_a_POST_request(self):
...
-    
+
     self.assertTemplateUsed(response, "home.html")
 ----
 ====
 ////
-// JAN: Such test will pass as long as you use something to persist TODO items across requests. To make it pass at this point, you'll need to either spent a lot of code to trick it or use models which is simpler.
+
+// JAN: Such test will pass as long as you use something to persist TODO items across requests.
+// To make it pass at this point, you'll need to either spent a lot of code to trick it or use models which is simpler.
```
hjwp, line 1216: hey @jangia just to let you know i'm still v intrigued by this suggestion.  i had a first go at it, see notes below, it's not quite working yet but i hope to be able to try again at some point.

one weird thing is that it's actually quite easy to persist things across requests in your tests, you just use a global variable.  and that actually  *more powerful* persistence than django models as far as tests are concerned, because django resets the db between each test, whereas the global variable sits there polluting state across tests for the whole session lol.

so haven't quite found the right approach to the chapter yet but i'll keep trying.

hjwp:
thought i'd merge this for now and continue working on things in a new PR

PR #179, opened 2023-07-19
authenticate method now requires request parameter

jwm-evans:
As far back as Django 1.11, it is necessary to pass the request object to the authenticate method otherwise the method does not match the signature of the parent and does not get called. For reference: https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#writing-an-authentication-backend

This remains true for Django 4.2


jwm-evans:
I've played around with it a bit more, and by changing the declaration of authenticate to :
`def authenticate(self, request=None, uid=None)`, the call to authenticate in accounts/views.py can remain as 
`user = authenticate(uid=uid)`

PR #180, opened 2023-07-19
authenticate method now requires request parameter 

jwm-evans:
As of Django 2.1, it is necessary to have the request object in the authenticate method signature otherwise the method does not  get called. For reference: https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#writing-an-authentication-backend

This remains true for Django 4.2


hjwp:
will do this one in 3e as well.  thanks!

PR #181, opened 2023-07-20
Review  chapter working incrementally

Xronophobe:
Marathon chapter, indeed! :) I wouldn't split it up though because the sub-titles help quite a lot with making sensible checkpoints for breaks during reading.
I especially liked when you only shown the expected error messages from the failing tests under Biting the Bullet: Adjusting Our Models.

What do you think about adding an Appendix: Glossary for terms like "Three strikes and refactor" or "Regression test", "YAGNI"?

### DIFF
--- a/chapter_working_incrementally.asciidoc
+++ b/chapter_working_incrementally.asciidoc
@@ -1336,7 +1336,7 @@ Again, a diff is a good way to see the changes:
 ====
 
 We create a new `List` object
-and then we assign each item to it by assigning it as its `.list` property.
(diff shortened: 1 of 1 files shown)


chapter_working_incrementally.asciidoc
```diff
@@ -1165,7 +1176,8 @@ def home_page(request):
 ====
 //24
 
-Yep!
+According to our unit tests: Yep!
```
hjwp, line 1179:
```suggestion
Yep!
```

PR #182, opened 2023-07-25
Review of chapter chapter_working_incrementally.asciidoc

Enforcer:
Another round of review from me.

Really liked this chapter as it lets to experience what it feels like to be a TDDer first-hand.

Added just a few slight comments to provide more context in one or two places. The only constructive feedback I have is for testing context passed to `render` function.

### DIFF
--- a/chapter_working_incrementally.asciidoc
+++ b/chapter_working_incrementally.asciidoc
@@ -61,6 +61,10 @@ it sounds like we're going to store lists and their items in a database.
 Each list will have a unique URL,
 and each list item will be a bit of descriptive text, associated with a particular list.
 
+// SEBASTIAN: Maybe adding some simple diagram showing dependency between lists, items and users might be helpful?
+//      Some people think visually, and showing where we were and where are we going to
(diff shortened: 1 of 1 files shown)


chapter_working_incrementally.asciidoc
```diff
@@ -73,7 +77,7 @@ we might want to recognise users using usernames and passwords,
 we might want to add a longer notes field as well as short descriptions to our list,
 we might want to store some kind of ordering, and so on.
 But we obey another tenet of the agile gospel:  "YAGNI" (pronounced yag-knee),
-which stands for "You ain't gonna need it!"
+which stands for "You ain't gonna need it!".
```
hjwp, line 80: i think the exclamation mark counts as sufficient punctuation to end the sentence.  oreilly also have the strange american convention of putting closing punctuation _inside_ quotes or brackets, even if they don't belong there.  it's a USA thing...

hjwp:
```suggestion
which stands for "You ain't gonna need it!"
```

PR #183, opened 2023-07-25
Update logout view

jwm-evans:
django.contrib.auth.views.logout seems to have been replaced by  django.contrib.auth.views.logout_then_login, which has a different parameter name (login_url instead of next_page). I'm not sure when this change was made


hjwp:
looking at this as part of the 3e.  thanks James!

PR #189, opened 2023-08-22
Review working incrementally

seddonym:
I enjoyed this chapter very much, but I found myself disagreeing a bit more with some of the approaches. All opinion-based though!

In particular, I feel like we're rather divorced from actually using the application. I have seen developers focus too much on the tests and never actually try using what they're building, and they end up shipping things that don't actually work. IMO it would be helpful to encourage the occasional return to trying it out as a user, especially when debugging FTs.

If we were to do that, as I was, we would need to address handling the addition of the non-nullable foreign key to the model. I ran into problems running migrations in my local database because there were already items in there. Have left a comment with a suggestion on that.

I also feel that some of this code isn't the best Django. For example, the URL scheme would be more conventional if we had trailing slashes everywhere and kebab case. It's also best practice to use reverse lookups instead of hardcoding the URLs, e.g. `reverse("view_list", item.id)`. But perhaps all that is coming later. I don't feel really strongly, though - I do get that maybe there is an advantage in not overburdening the reader with Django stuff. It's just that if you do know Django, this code feels a bit below par, and maybe we don't want that.

Finally - personally, some of these tests feel too coupled to implementation. I do get that we are following a discipline to an extreme here, but my understanding of TDD is that we ideally don't want the tests to get in the way of refactoring, and in some cases they do here. I have definitely had to learn to write less tests, a few years ago I was doing more of this kind of thing (every function needed a test) and it took time to learn to relax that a bit.

PS I seem to have broken the tests, not sure why or if this matters, let me know if you'd like me to do anything differently.

### DIFF
--- a/chapter_07_working_incrementally.asciidoc
+++ b/chapter_07_working_incrementally.asciidoc
@@ -43,6 +43,8 @@ Instead of a long up-front design phase,
 we try to put a _minimum viable application_ out there early,
 and let the design evolve gradually based on feedback from real-world usage.
 
+// DAVID: I would say the more common term is Minimum Viable Product, I haven't heard people say
+// 'minimum viable application.'
 
 But that doesn't mean that thinking about design is outright banned!
 In the last big chapter we saw how just blundering ahead without thinking can _eventually_ get us to the right answer,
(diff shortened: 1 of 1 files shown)


hjwp:
thanks david!

had to resolve some merge conflicts, most of which were due to the fact that @Enforcer had made similar or related comments in many of the same places you did!  nice to see some consensus.

i think there's also some consensus that the book has too much "testing the implementation" and overly coupled tests, so I want to do something about that.  will start a chat on the group.

don't worry too much about breaking the tesst, but basically, any changes to the source code listings will break the tests because they are all tested both functionally and by a big git diff at the end...  by all means continue to make suggestions in code tho!

PR #194, opened 2023-10-16
Review chapter_08_prettification.asciidoc

jangia:
- Suggest to mention force refresh for browser
- Suggest to briefly describe StaticLiveServerTestCase


hjwp:
good suggestions thanks!

PR #196, opened 2023-10-20
Review  chapter 08 prettification

Xronophobe:
Nice, faster-paced chapter, easy to follow and seeing the visual upgrade on the Lists page is a very rewarding feeling.

I was wondering whether Bootstrap is still a future-proof choice, considering the changes at Twitter, but it seems like it is: https://github.com/orgs/twbs/discussions/38986

ps.: I just noticed we are going with Python 3.12 now. I was using 3.11 for reviewing the chapter, but I think it shouldn't be a problem.


chapter_08_prettification.asciidoc
```diff
@@ -971,7 +1012,7 @@ INSTALLED_APPS = [
     "django.contrib.sessions",
     "django.contrib.messages",
     "django.contrib.staticfiles",
-    "lists",
+    "lists.apps.ListsConfig",
```
Xronophobe, line 1015: Btw my `INSTALLED_APPS` have single quote entries, why are these double quotes? Maybe some auto-formatter?

chapter_08_prettification.asciidoc
```diff
@@ -822,6 +859,7 @@ All that took me a few goes, but I'm reasonably happy with it now
 [[homepage-looking-better]]
 .The lists page, looking good enough for now...
 image::images/prettified-final.png["Screenshot of lists page in light mode with decent styling."]
+// CSANAD:  This image seemed to be broken, so I uploaded a new one.
```
Xronophobe, line 862: Please check the resolution, my image ended up significantly smaller in size.

chapter_08_prettification.asciidoc
```diff
@@ -20,8 +20,8 @@ Our site is undeniably a bit unattractive at the moment
 
 
 NOTE: If you spin up your dev server with `manage.py runserver`,
-    you may run into a database error
-    "table lists_item has no column named list_id".
+    you may run into an operational error
+    "no such table: lists_list".
```
hjwp, line 24:
```suggestion
    you may run into a database error, something like this:
    "table lists_item has no column named list_id".
```

chapter_08_prettification.asciidoc
hjwp: i ran black on *everything* at some point.  and i think django does too now, so everything in settings.py has double quotes these days...

(also i change to just using "lists" rather than `lists.apps.ListConfig` because the latter just seems needlesscly complicated.

chapter_08_prettification.asciidoc
```diff
@@ -975,7 +1016,7 @@ INSTALLED_APPS = [
     "django.contrib.sessions",
     "django.contrib.messages",
     "django.contrib.staticfiles",
-    "lists",
+    "lists.apps.ListsConfig",
```
hjwp, line 1019:
```suggestion
    "lists",
```

chapter_08_prettification.asciidoc
hjwp: thanks Csanad!

chapter_08_prettification.asciidoc
Xronophobe: Oh, I didn't notice this change. Got it.
https://github.com/hjwp/Book-TDD-Web-Dev-Python/commit/8c19394faf006933ff549ecb5f1bee943b0990a4#diff-7beedde00b2fc5c59611d91793c5cec65512ef6f7202caaa18980032a7ad4c1bR513

PR #199, opened 2023-10-25
Review prettification

seddonym:
I enjoyed!

There are probably some things that I would do differently. I think starting with Bootstrap's CDN so they get a nicely looking site really quickly, and then adding a custom `style.css` with collectstatic might be better, as it doesn't involve faffing about with so many files. But not a big deal.

I'm also not convinced this is really how people should test rendering in real life. We don't do anything like this, but if we were to then I would say some mechanism for comparing rendered screenshots would work better. Or, if we want to do something simple, just something to check the statics don't 404. 🤷🏻


hjwp:
> some mechanism for comparing rendered screenshots would work better. 

yes, i've used those in the past.  i think part of my original motivation was to show that selenium has some methods for inspecting style + layout properties, like x/y coordinateas, which can come in useful sometimes.  i've used them to test some fiddly alignment / window resizing behaviours...   thanks, i'll mention that explicitly i think.

PR #200, opened 2023-11-02
Update chapter_12_organising_test_files.asciidoc

jjasghar:
The "fn" isn't used in this function, but "row_text" is.

### DIFF
--- a/chapter_12_organising_test_files.asciidoc
+++ b/chapter_12_organising_test_files.asciidoc
@@ -491,7 +491,7 @@ and we'll adapt it slightly:
 ====
 [source,python]
 ----
-    def wait_for(self, fn):  #<1>
+    def wait_for(self, row_text):  #<1>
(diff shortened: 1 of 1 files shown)


jjasghar:
Oh dang, i miss read it. Sorry about that.

hjwp:
thanks so much for submitting a PR tho!

those listings are a little confusing maybe, because i break down creating the method into two steps...

in other news -- did you have a go at the deployment?  would you like to beta test my new versions of the deployment chapters?  they've got docker in...

jjasghar:
@hjwp I'd love to beta test your deployment. Are you just doing local docker-compose or something like kubernetes/openshift?

No joke, because of your book I have able to get a simple django app deployed and ran on OpenShift for my job. (I had problems with the static stuff, turns out your book had the _correct answer_ for me, so I started at the beginning and loving every min of it :) )

PR #201, opened 2024-01-22
Update chapter_09_docker.asciidoc

nicknielsen:
Fixed a few typos and made some stylistic suggestions in the beginning of Chapter 9.

### DIFF
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -261,7 +261,7 @@ But there are solutions to all of these.  In order:
 
 === An Overview of Our Deployment Procedure
 
-Over these three chapters chapters I'm going to go through _a_ deployment procedure.
+Over these three chapters, I'm going to go through _a_ deployment procedure.
 It isn't meant to be the _perfect_ deployment procedure,
 so please don't take it as being best practice,
 or a recommendation--it's meant to be an illustration,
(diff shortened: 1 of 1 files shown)


hjwp:
thanks very much!  typo corrections always much appreciated.

someone at oreilly will be in charge of double-checking all the TitleCasing of section headers when we go to production, but it does absolutely no harm to try and get them right in advance, so those are very much appreciated too!

any general thoughts on the chapter?  did you manage to follow along with everything ok?

nicknielsen:
@hjwp No problem at all! Happy to help wherever I can.

I'm really liking the book. It's opened my eyes to whole new way of programming and thinking about my projects. I actually just came back to it yesterday after a little holiday-related break, so I was pretty surprised to see all the new content about containers!

Before I set the book down for the holidays, I was a little over halfway through Chapter 9. At that time, the chapter still had me pushing the codebase directly to the staging server first. I'm going to start the chapter over with the new method, and I'll let you know if I run into any snags as I make my way through it.

hjwp:
Brilliant thanks!  The next chapter is pretty much there, definitley would appreciate feedback on that. and then chapter 11 on how to actually deploy to a server with ansible is VERY rough so sorry about that in advance, but let me know how you get on if you do decide to give it a go!

after that chapters 12-14 have been upgraded to django 4 but i haven't reviewed the actual text content yet, and chapters 15+ are on django 1.x...

PR #202, opened 2024-01-23
review chapter 09

Xronophobe:
I enjoyed this chapter and I agree with the approach and the steps. Later on I'm hoping to see some Dockerfile optimizations and using a dedicated user instead of root, but for now it's completely sufficient.


chapter_09_docker.asciidoc
```diff
@@ -778,6 +823,14 @@ TIP: Forgetting the `-r` and running `pip install requirements.txt`
     because (at the time of writing), it's not very self-explanatory.
     And it's a mistake I still make, _all the time_.
 
+// CSANAD: This is no longer true, the error message is very helpful by now:
```
hjwp, line 826: TIL!  hooray!  <3

hjwp:
thanks so much for this Csanad!  your attention to detail is a great help! ❤️

PR #206, opened 2024-02-19
fix: chapter 2 typos

awdem:
1) adds a missing 'out' 
2) correct 'Philosohpy' to 'Philosophy'


hjwp:
Thanks very much Will!  Let me know how you get on with the book!  Particularly if/when you get to chapter 9 (assuming you're reading the 3e / online / early release?)

PR #207, opened 2024-02-27
Rita edits to ch 9

ritaf-ORM:


chapter_09_docker.asciidoc
```diff
@@ -1,5 +1,6 @@
 [[chapter_09_docker]]
 == Deployment Part 1: Containerization aka Docker
+// RITA: I'm not keen on including the word "part" in chapter titles especially when the chapter is also within something called a "part." Would it make sense to put the deployment chapters into its own Part? So, ch8 might go into Part 1, chs 9-11 would go into the new Part 2 called something like "Deployment," and Part 3 would start with ch 12.
```
hjwp, line 3: great suggestion!  i think i'm still resistant to change the parts - part 1 being the basics of tdd, part 2 being "other web dev stuff", and part 3 being "more advanced tdd", i think those groupings still make sense.

but you're definitely right that there's something missing here in the way i'm grouping the 3 "deployment" chapters.  i'll have a think about it.

chapter 8 really doesn't feel like it goes with part 1, you see...

(just thinking out loud now) 12 could definitely go into part 1, at least conceptually...

13/14/15 also all kind of go together.

you need chapter 8 before you can do chapters 9-11.

no easy solution i can see!

for now i'll just strip out the "part x" frlm the 3 deployment chapters...

chapter_09_docker.asciidoc
```diff
@@ -927,7 +928,7 @@ WORKDIR /src
 CMD python manage.py runserver 8888
 ----
 ====
-
+//RITA: Newbie question. What does Ctrl+C do again? Add a couple of words to give us some quick context. "Use Ctrl+C to kill the current..."
```
hjwp, line 931: ctrl+c is so common in techie circles i find it hard to imagine anyone that hasn't come across it yet!  maybe if this was a pure beginners book...

chapter_09_docker.asciidoc
```diff
@@ -1391,7 +1394,8 @@ Ran 3 tests in 26.965s
 OK
 ----
 
-AMAZING IT ACTUALLY WORKSSSSSSSS.
+//RITA: I'd rather add exclamation marks than extend the word.
```
hjwp, line 1397: hope it's ok if this is one of the few we keep!

PR #210, opened 2024-03-11
review chapter09 (rev2)

Xronophobe:
As Rita asked, I had another look on this chapter and found a few more minor things. I was trying not to repeat myself but I just noticed I did with the non-privileged user in the Docker container - suggestion. 
Overall, I like the improvements!

### DIFF
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -857,6 +857,16 @@ WARNING: Make sure you use the `-it` flags to the Docker `run`
 *******************************************************************************
 If you've got a container that's "hanging" in a terminal window,
 you can kill it from another one.
+// CSANAD:  `docker stop` would trigger a graceful SIGTERM, `docker kill` sends a SIGKILL.
+//          SIGKILL would terminate the container even when SIGTERM cannot, but it could
+// also result in broken DB, leave the port "in use" or other side effects. I would prefer
+// using `docker stop` over `kill` and mention `kill` in case the container is still stuck.
+//
+// However, if we leave it with just `docker kill`, then maybe we should mention these
+// possible side effects and also the title should reflect it like:
+// "How to Forcefully Stop a Docker Container" (since "How to Kill a Container" may sound a bit
+// harsh...
+
 
 The docker daemon lets you list all the currently running containers
 with `docker ps`:
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -1091,8 +1091,10 @@ But we're just mapping 8888 to 8888, and that will look like this:
 
 [subs="specialcharacters,quotes"]
 ----
-$ *docker build -t superlists . && docker run -p 8888:8888 -it superlists*
+$ *docker run -p 8888:8888 -it superlists*
 ----
+
+TIP:  We don't need to run `docker build` this time since we have not changed our Dockerfile.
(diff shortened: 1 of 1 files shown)


chapter_09_docker.asciidoc
```diff
@@ -102,7 +100,7 @@ Security and Configuration::
 One way to approach the problem is to get a server,
 and start manually configuring and installing everything,
 hacking about until it works,
-and maybe think about automating things laterfootnote:[
```
hjwp, line 105: i know it looks nicer in the asciidoc source if you put the full stop before the `footnote`, but it looks worse in print, so someone will change it back during copyediting...

chapter_09_docker.asciidoc
```diff
@@ -58,8 +58,6 @@ Let me know how you got on, if you get stuck on anything,
 if any explanations don't make sense,
 or if any of the instructions don't work for you.
 
-I hope you enjoy the new version!
```
Xronophobe, line 61: I'm not sure why this line shows as deleted. I just fetched and rebased from upstream. I'll have another look soon/

Xronophobe: I must have had it open in vim while I fetched and rebased but then didn't reload the file, so I just saved it without those lines. Fixed now

chapter_09_docker.asciidoc
```diff
@@ -1039,6 +1061,15 @@ HTTP/1.1 200 OK
 </html>
 ----
 
+// CSANAD: I think the readers' inputs were a little difficult to spot, so I added some in-
```
Xronophobe, line 1064: actually, the bold text might be enough after all

chapter_09_docker.asciidoc
```diff
@@ -1064,9 +1095,11 @@ But we're just mapping 8888 to 8888, and that will look like this:
 
 [subs="specialcharacters,quotes"]
 ----
-$ *docker build -t superlists . && docker run -p 8888:8888 -it superlists*
+$ *docker run -p 8888:8888 -it superlists*
 ----
 
+TIP:  We don't need to run `docker build` this time since we have not changed our Dockerfile.
```
Xronophobe, line 1101: I just found this. Now that we had a call with Harry recently that made it clear we were not mounting src/ in the next chapter on purpose, this is just partially true. 
Correctly: 

> "We don't need to run `docker build` this time since we are not changing our image."

The way I worded it originally suggests that we only need to rebuild when we change the Dockerfile itself (which is usually the case with most of my work) but since we use the COPY directive in the Dockerfile to bake the code in the image, modifying the source code also only takes effect after a rebuild is done.

I'll commit the change soon.

chapter_09_docker.asciidoc
```diff
@@ -388,6 +388,10 @@ TIP: I'm deliberately choosing a different port to run Dockerised Django on (888
 
 I'll use the `--failfast` option to exit as soon as a single test fails:
 
+// CSANAD:  are line breaks necessary below and at other occurrences of
+//          --failfast? it does keep the asciidoc lines of source below 80
+// characters but there are some wider lines elsewhere.
```
hjwp, line 393: by default, source code has to break around 80 chars, otherwise it won't fit on the printed page.

there is a `role=small-code` directive we can use when necessary, but i try and avoid it when i can...

PR #214, opened 2024-03-19
Review chapter 10

seddonym:
Solid chapter. I'm glad you broke this out from the previous one, that one had quite a lot to digest and this is a bit easier.


hjwp:
thanks very much David!  thorough and thoughtful comments as always.  i really appreciate you putting in the time to go through it manually.

PR #217, opened 2024-03-28
Review chapter_10_production_readiness.asciidoc

jangia:
One general comment: The Dockerfile at the end of the chapter is far from production-ready. e.g., the root user is used to run the server, etc. I'd suggest you check [Dockerizing Django with Postgres, Gunicorn, and Nginx](https://testdriven.io/blog/dockerizing-django-with-postgres-gunicorn-and-nginx/) and [Docker Best Practices for Python Developers](https://testdriven.io/blog/docker-best-practices/#prefer-array-over-string-syntax). I don't think you should go into details, but IMO it would be beneficial to have an actual production-ready Docker image at the end of the chapter. People usually take things they find in courses and books as they are and use them in production. Especially the deployment-related things - "It works. Let's just use it".

### DIFF
--- a/chapter_10_production_readiness.asciidoc
+++ b/chapter_10_production_readiness.asciidoc
@@ -185,7 +185,7 @@ $ *docker build -t superlists . && docker run \
 ----
 
 And if you take another manual look at your site, things should look much healthier.
-Let's rerun our FTs to confirm:
+Let's rerun our FTs to confirm:  // JAN: That doesn't work until you install whitenoise to venv on local machine
 
 [role="small-code"]
 [subs="specialcharacters,macros"]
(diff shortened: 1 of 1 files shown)


hjwp:
thanks for the suggestions Jan!  I'll try and take on a couple of extra production-readiness ideas.  switching to a nonroot user seems like a good one, and probably the array syntax for CMD too.  Any others you think are "must-haves"?

I don't think i want ot get into caching or multistage builds, or adding nginx/docker-compose for example.  but if there are any other key things you think it might be nice to show in examples, rather than suggest in terms of further reading (i can definitely link to those posts on testdriven.io for example)  then lnk.

jangia:
Sounds reasonable to me👍

PR #218, opened 2024-04-02
Update chapter 12

Xronophobe:
I ended up doing both an update and review of Chapter 12. I found all code examples were up to date and working and I couldn't find any best practices or tips that were outdated either.
I added a few suggestions and edits. 

I ended up not committing a comment at line 669, but I still wanted mention that adding new lines describing further test scenarios before we make the first one pass feels less ~~right~~ comfortable. But it isn't bad, just something I noticed.

It was a faster, shorter chapter; easy to follow and it was nice to see the growing test files end up split up into their corresponding files.

### DIFF
--- a/chapter_12_organising_test_files.asciidoc
+++ b/chapter_12_organising_test_files.asciidoc
@@ -159,7 +159,7 @@ and wait until we're back to a fully passing test suite before refactoring.
 
 Kent Beck has a book-length exploration of the tradeoffs
 of refactor-now vs refactor-later, called
-https://learning.oreilly.com/library/view/tidy-first/9781098151232/[Tidy First?]
+https://www.oreilly.com/library/view/tidy-first/9781098151232/[Tidy First?]
 **********************************************************************
 
 
@@ -291,7 +291,7 @@ Some people like to use them a lot in Django code
 (e.g., your views might import models using `from .models import List`,
 instead of `from list.models`).
 Ultimately this is a matter of personal preference.
-I prefer to use relative imports only when I'm super-super
+I prefer to use relative imports only when I'm super-super confident
 that the relative position of the thing I'm importing won't change.
 That applies in this case because I know for sure
 all the tests will sit next to _base.py_, which they inherit from.
@@ -385,7 +385,7 @@ For now, let's commit!
 ----
 $ *git status*
 $ *git add src/functional_tests*
-$ *git commit -m "Moved Fts into their own individual files"*
+$ *git commit -m "Moved FTs into their own individual files"*
 ----
 
 Great.  We've split our functional tests nicely out into different files.
@@ -631,7 +631,7 @@ through what happened:
 <3> To explain where the exception has actually come from,
     the traceback takes us back into _test_list_item_validation.py_
     and inside the body of the `lambda` function,
-    and tells us that it the attempt to find the `.invalid-feedback` element
+    and tells us that the attempt to find the `.invalid-feedback` element
     that failed.
 
 
--- a/chapter_12_organising_test_files.asciidoc
+++ b/chapter_12_organising_test_files.asciidoc
@@ -381,9 +381,10 @@ AssertionError: write me!
 Brilliant--no need to sit around waiting for all the FTs
 when we're only interested in a single one.
 Although we need to remember to run all of them now and again, to check for regressions.
-It's a good habit to run all the tests before making a `git commit`.
-Later in the book we'll see how to give that task over to an automated Continuous Integration loop.
-For now, let's commit!
+Later in the book we'll set up a Continuous Integration (CI) server to run all the tests automatically,
+for example every time we push to master.
+For now, a good prompt for running all the tests is "just before you do a commit",
+so let's get into that habit now:
 
 [subs="specialcharacters,quotes"]
 ----
--- a/chapter_12_organising_test_files.asciidoc
+++ b/chapter_12_organising_test_files.asciidoc
@@ -381,6 +381,7 @@ AssertionError: write me!
 Brilliant--no need to sit around waiting for all the FTs
 when we're only interested in a single one.
 Although we need to remember to run all of them now and again, to check for regressions.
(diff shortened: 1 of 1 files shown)


chapter_12_organising_test_files.asciidoc
```diff
@@ -378,14 +381,15 @@ AssertionError: write me!
 Brilliant--no need to sit around waiting for all the FTs
 when we're only interested in a single one.
 Although we need to remember to run all of them now and again, to check for regressions.
+It's a good habit to run all the tests before making a `git commit`.
 Later in the book we'll see how to give that task over to an automated Continuous Integration loop.
 For now, let's commit!
```
hjwp, line 386: i'm worried this ends up being ambiguous - are we giving the task of making commits over to CI?

```suggestion
Although we need to remember to run all of them now and again, to check for regressions.
Later in the book we'll set up a Continuous Integration (CI) server to run all the tests automatically,
for example every time we push to master.
For now, a good prompt for running all the tests is "just before you do a commit",
so let's get into that habit now:
```

chapter_12_organising_test_files.asciidoc
```diff
@@ -547,6 +551,8 @@ Like this:
 ((("Python 3", "lambda functions")))
 `lambda` in Python is the syntax for making a one-line, throwaway function--it
 saves you from having to use `def..():` and an indented block:
+// CSANAD: the `def..` and the `():` renders apart from each other, on two
+//         lines. I couldn't find out how to keep them together.
```
hjwp, line 555: there's a special role thingie called `keep-together` - syntax is something like this,
but it might not render quite right if we use it at the same time as backticks:

```suggestion
saves you from having to use [keep-together]#`def..():`# and an indented block:.
```

could also try just `[keep-together]`def..():` hmmm

chapter_12_organising_test_files.asciidoc
Xronophobe: Nice, I like it!

hjwp: maybe we'd want to mark it as xfail then?

chapter_12_organising_test_files.asciidoc
Xronophobe: Yeah, I tried that but didn't work. How about cheating a little with an extra word or two:
 "...from having to use **the regular** `def..():` and an indented block:" ?

chapter_12_organising_test_files.asciidoc
```diff
@@ -547,6 +552,8 @@ Like this:
 ((("Python 3", "lambda functions")))
 `lambda` in Python is the syntax for making a one-line, throwaway function--it
 saves you from having to use `def..():` and an indented block:
```
Xronophobe, line 554: Apparently I can't make my comment below a committable change, so I'll just suggest it again.
I tried the `keep-together` but it didn't work within a code block. How about cheating by adding a word:
```
saves you from having to use the regular `def..():` and an indented block:
```

chapter_12_organising_test_files.asciidoc
hjwp: just ignore it for now, they'll pick it up in copyediting.

chapter_12_organising_test_files.asciidoc
```diff
@@ -561,9 +568,10 @@ saves you from having to use `def..():` and an indented block:
 5
 ----
 
-In our case, we're using it to transform a bit of code that would otherwise be
-executed immediately into a function that we can pass as an argument,
+In our case, we're using it to transform a bit of code - that would otherwise be
+executed immediately - into a function that we can pass as an argument,
 and that can be executed later, and multiple times:
+// CSANAD: I think it's easier to read like this.
```
hjwp, line 574: not sure i agree but nbd.

fwiw the preferred syntax for emdashes is `word--nextword`

chapter_12_organising_test_files.asciidoc
```diff
@@ -561,9 +568,10 @@ saves you from having to use `def..():` and an indented block:
 5
 ----
 
-In our case, we're using it to transform a bit of code that would otherwise be
-executed immediately into a function that we can pass as an argument,
+In our case, we're using it to transform a bit of code - that would otherwise be
+executed immediately - into a function that we can pass as an argument,
```
hjwp, line 572:
```suggestion
In our case, we're using it to transform a bit of code--that would otherwise be
executed immediately--into a function that we can pass as an argument,
```

chapter_12_organising_test_files.asciidoc
```diff
@@ -621,18 +629,24 @@ FAILED (errors=1)
...
 
-<1> At line 15 in our FT, we go into our `self.wait_for` helper,
-    passing it the `lambda`-ified version of the `assertEqual`.
+<1> In our FT, we call our `self.wait_for` helper, where we pass
+    the `lambda`-ified version of `assertEqual`.
 
-<2> We go into `self.wait_for` in _base.py_,
-    where we can see that we're inside `fn()`,
-    which is the name for our lambda inside the helper.
+<2> We go into `self.wait_for` in `base.py`,
```
hjwp, line 635: oreilly convention is that filenames are in italics 

```suggestion
<2> We go into `self.wait_for` in _base.py_,
```

chapter_12_organising_test_files.asciidoc
```diff
@@ -621,18 +629,24 @@ FAILED (errors=1)
...
 <3> To explain where the exception has actually come from,
     the traceback takes us back into _test_list_item_validation.py_
     and inside the body of the `lambda` function,
-    and tells us that it the attempt to find the `.invalid-feedback` element
+    and tells us that the attempt to find the `.invalid-feedback` element
     that failed.
+// CSANAD: I changed two things here:
+//   First, I think it's better to remove the specific line number which may
+// vary due to personal formatting preferences (and following the pep8 puts
+// this on line 16 for me)
```
hjwp, line 647: 💯

hjwp (approved):

PR #220, opened 2024-04-03
Review of chapter 10

Enforcer:
This chapter was much lighter than chapter 9th, but still very relevant.

Those summaries, especially "Production-Readiness Config" are great. They provide a checklist-like guide for people not to get lost and are nice to see at the end of the chapter.

I am not very fond of freezing just "top-level" dependencies of the project, as it is shown in the chapter. This is not guaranteeing the reproducible builds and I failed to build & deploy because of that more than once. But I understand it's a trade-off so the reader does get bogged down.

Apart from that, one suggestion about linking Django docs page for deployment for curious readers.

### DIFF
--- a/chapter_10_production_readiness.asciidoc
+++ b/chapter_10_production_readiness.asciidoc
@@ -51,6 +51,9 @@ and as we'll see, other things will come up.
 
 Let's go through and see if we can fix things one by one.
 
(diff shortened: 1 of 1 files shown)


hjwp:
and thanks again!  i'll add a little reminder to myself about the top-level package pinning.

PR #223, opened 2024-04-12
Chapter 11 review

jangia:
Added some comments. The main issue is that I was stuck with Docker installation on the server via Ansible. After checking it on Google, it seems like much more is required to successfully install Docker (like adding certs, ...).


chapter_11_ansible.asciidoc
```diff
@@ -318,6 +324,10 @@ for.
 ----
 ====
 
+// JAN: I'm stuck here. Docker is not installed on EC2 instance with Ubuntu. Googling it, it seems like this is some outdated Docker installation process
+The one described here is much closer to Docker's installation guide: https://www.digitalocean.com/community/tutorials/how-to-use-ansible-to-install-and-set-up-docker-on-ubuntu-22-04
```
hjwp, line 328: hi Jan, sorry about this, my fault, I've since updated the book to say you need to install `docker.io` not `docker`

(you can also do a bunch of faffing with docker's private package store but i want to avoid that for the book if i can)

Issue #226, opened 2024-04-14
Chapter 10: callout numbers for annotation texts are not correct in source code block ch09l005 


paatre:
`asciidoctor` will warn of the following errors when `make chapter_10_production_readiness.html` is ran:

```sh
asciidoctor: WARNING: chapter_10_production_readiness.asciidoc: line 374: callout list item index: expected 1, got 3
asciidoctor: WARNING: chapter_10_production_readiness.asciidoc: line 376: callout list item index: expected 2, got 4
```

When a source block is annotated by callout numbers, same numbers should be used in the annotation text section. But this is not the case in `chapter_10_production_readiness.asciidoc`:

```asciidoc
.Dockerfile (ch09l005)
====
[source,dockerfile]
----
FROM python:slim

RUN python -m venv /venv
ENV PATH="/venv/bin:$PATH"

COPY requirements.txt requirements.txt  <1>
RUN pip install -r requirements.txt  <2>

COPY src /src

WORKDIR /src

CMD python manage.py runserver
----
====

<3> We COPY our requirements file in, just like the src folder.

<4> Now instead of just installing Django, we install all our dependencies
  by pointing pip at the _requirements.txt_ using the `-r` flag.
  Notice the `-r`.
```

### Changelist

- [ ] Change `<3> We COPY our requirements file in...` into `<1> We COPY our requirements file in...`
- [ ] Change `<4> Now instead of just installing Django...` into `<2> Now instead of just installing Django...`

paatre:
These have been merged in #227.

hjwp:
thanks Teemu!

Issue #225, opened 2024-04-14
Chapter 10: Django 4.2.7 still displayed as the result of pip freeze | grep -i django


paatre:
In `chapter_10_production_readiness.asciidoc`, when the `requirements.txt` is created by adding, there's still an old Django version displayed even though in the next `sourcecode` section, the Django version is `4.2.11`:

```
[subs="specialcharacters,quotes"]
----
$ *pip freeze | grep -i django*
Django==4.2.7 << HERE

$ *pip freeze | grep -i django >> requirements.txt*
$ *pip freeze | grep -i gunicorn >> requirements.txt*
$ *pip freeze | grep -i whitenoise >> requirements.txt*
----
```

hjwp:
thanks!  I've fixed that 😊

PR #227, opened 2024-04-14
Minor chapter 10 asciidoc fixes

paatre:
There are three fixes on this PR:

1. #225
2. #226 
3. I also noticed `ch10l005` commit ref was actually written as `ch09l005` which is now fixed


hjwp:
thanks very much indeed Teemu!

PR #231, opened 2024-04-22
Update and review Chapter 13

Xronophobe:
I couldn't find any outdated methods or practices in the chapter. 
The use of `$ python ...` and `$ python3 ...` are a little inconsistent.
Also, in earlier chapters I think we were being more specific about which test to run even if it was clear from the output attached below the text. In this chapter we often don't do this, rather just say "the test". On the other hand, it can be expected from the reader to get a feel for what the book is aiming to teach by now, which would mean fewer very specific step-by-step instructions, so I don't have a strong opinion on this one.
 
Other than these, I found it easy to read and satisfying to see how the code is becoming more concise and pretty.

### DIFF
--- a/chapter_13_database_layer_validation.asciidoc
+++ b/chapter_13_database_layer_validation.asciidoc
@@ -115,7 +115,7 @@ TIP: If you're new to Python, you may never have seen the `with` statement.
     It's used with what are called "context managers", which wrap a block of code,
     usually with some kind of setup, cleanup, or error-handling code.
     There's a good write-up in the
-    http://docs.python.org/release/2.5/whatsnew/pep-343.html[Python 2.5 release notes].
+    https://docs.python.org/release/2.5/whatsnew/pep-343.html[Python 2.5 release notes].
     ((("with statements")))
     ((("Python 3", "with statements")))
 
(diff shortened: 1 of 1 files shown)


chapter_13_database_layer_validation.asciidoc
```diff
@@ -187,11 +198,17 @@ Let's hack it in to see it work:
...
+        item.save()
 ----
 ====
 //19
+// CSANAD: full_clean() should be executed before save() is called. It doesn't
+//         make a difference here since the test DB is destroyed anyway, but
+// we are actually saving an invalid item and only then calling full_clean().
+// Which results in raising the correct error. But I think we should always
+// show the correct order even when it doesn't matter: validation first and
+// only then attempting save().
```
hjwp, line 211: yep makes absolute sense.  i wonder why i had them in that order...

hjwp:
thanks very much Csanad, great suggestions.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -241,8 +258,15 @@ been passed an error variable, and if so, we do this:
...
 ((("form control classes (Bootstrap)")))
-Take a look at the http://getbootstrap.com/css/#forms[Bootstrap docs] for more
+Take a look at the https://getbootstrap.com/docs/5.3/forms/validation/#server-side[Bootstrap docs] for more
 info on form controls.
+// CSANAD: these are the new docs for Bootstrap, but for some reason they begin
+//         with saying "We recommend client-side validation" which is bad.
+// Client side validation is fine for faster UI aesthetics, sure. But they are
+// not emphasizing that it is only for faster feedback and for security,
+// whatever reaches the server should be validated on the server side before
+// accepting it as an input for literally anything... Maybe we should mention
```
hjwp, line 268: maybe just a simple footnote "... and ignore their advice to prefer client-side validation.  (a) both is better and (b) they're going to say that because they're frontend people and (c) honestly if you cant do both then server-side validation is probably the one you really can't do without"

chapter_13_database_layer_validation.asciidoc
```diff
@@ -573,6 +599,19 @@ The current situation is that we have one view and URL for displaying a list,
...
+// And then just re-reference it from where it is right now.
+//
+// I think adding a few words on how this is not just a simple refactor but it's
+// going to get us to our working state would help the reader recognise these
+// situations later, independently.
+// Something like pointing out how the FT is partially passing for:
+//  - adding one empty list item results in an error message
+//  - adding one valid list item is also passing
+//  - the error occurs when we try adding an empty list item again
+// from which we conclude why we are refactoring the views first
```
hjwp, line 614: i think calling this out is a good idea yes.  i think maybe the rigorous thing to do might be to add an early `return` to the FT, with a `# TODO` on it, so that it still runs right up to the known error, but we have green fts, then we do a commit, then we refactor, then we commit, then we undo the early return... something like that?  would you like to have a go at it?

PR #232, opened 2024-04-24
Review Ansible chapter

seddonym:
I like this chapter, as it touches on a few things I'm less comfortable with. There are a few things to be ironed out though.

The main challenge is that I found that when I ran into problems I lacked the mental model to troubleshoot - it's possible that others who don't have access to Harry(TM) will give up. I think talking through what Ansible is doing, maybe even a diagram at the beginning of the chapter to show what we're aiming for? In particular, it's difficult to understand where the Ansible error logs are coming from (i.e. local, server, container...)

I also think we're missing some stuff at the end about how all this might look as a development workflow. Maybe talk about setting up scripts (so we don't have to remember the ansible command?) And what about releasing to production? It doesn't need much, it just feels unfinished to me.

A few small things:

- I think you should make more of the fact that our functional tests can be run against a real website hosted elsewhere. The fact that we can do that was not obvious to me when we wrote those tests. Worth talking about a bit more?
- Shouldn't we commit our changes to Git at some point, as per the other chapters?
- Will we be returning to this again in the book? I'd like to know whether I can destroy my Digital Ocean droplet yet, don't want to get billed needlessly.

### DIFF
--- a/chapter_10_production_readiness.asciidoc
+++ b/chapter_10_production_readiness.asciidoc
@@ -555,6 +555,9 @@ It's not quite working yet!  Let's take a look manually: <<django-400-error>>.
 .An ugly 400 error
 image::images/twp2_1002.png["An unfriendly page showing 400 Bad Request"]
 
+// DAVID: noticed in passing that the screenshots in this chapter are from the hosted version
+// which isn't covered until next chapter.
+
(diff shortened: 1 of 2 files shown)


hjwp:
thanks david!  great comments as always.  i think a diagram really would be a good idea.

PR #233, opened 2024-05-03
chapters 8 and 9: propose a few small edits

zipperer:
This pull request proposes a few small edits for chapters 8 and 9. Each of the edits is independent of the others. I can edit/revert any commit deemed incorrect/unwarranted. 

For each commit, the commit message gives more info. 

Please share any questions, and it won't hurt my feelings if edits are denied.

### DIFF
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -496,7 +496,7 @@ Follow those, and then test your installation by running:
 
 [subs="specialcharacters,macros"]
 ----
-$ pass:quotes[*docker run busybox echo hello world*]
+$ pass:quotes[*docker run --rm busybox echo hello world*]
 Unable to find image 'busybox:latest' locally
 [...]
 latest: Pulling from library/busybox
@@ -512,6 +512,7 @@ What's happened there is that Docker has:
 * Downloaded the image from DockerHub
 * Created a container based on that image
 * Started up that container, telling it to run `echo hello world`
+* Deleted that container when the container stopped
 * And we can see it worked!
 
 Cool! We'll find out more about all of these steps as the chapter progresses.
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -956,7 +956,7 @@ Starting development server at http://127.0.0.1:8888/
 
 A quick run of the FT or check in our browser will show us that nope, that doesn't work either.
 Let's try an even lower-level smoke test, the traditional Unix utility `curl`.
-It's a command-line tool for making HTTP requests.  Try it on your own computer first:
+It's a command-line tool for making HTTP requests.footnote:[`curl` can make many other types of requests. For details see https://man7.org/linux/man-pages/man1/curl.1.html[curl manual]]  Try it on your own computer first:
 
 [role="ignore-errors"]
 [subs="specialcharacters,macros"]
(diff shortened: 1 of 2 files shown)


zipperer (commented):
53356dc938c03bcb3ccba5bdfc499013643ee483 reverts th change from b65eb467d9fb1f22db98675549217dcb0c78889a and adds `[subs="quotes"]` to make bold `docker ps` and `docker kill` inside the block listings.

chapter_09_docker.asciidoc
zipperer, line 1: This reverts the change from b65eb467d9fb1f22db98675549217dcb0c78889.

After this commit `docker ps` and `docker kill` are bold in the block listings:
chapter9_docker_ps_bold_in_listing_block

chapter_09_docker.asciidoc
```diff
@@ -496,7 +496,7 @@ Follow those, and then test your installation by running:
 
 [subs="specialcharacters,macros"]
 ----
-$ pass:quotes[*docker run busybox echo hello world*]
+$ pass:quotes[*docker run --rm busybox echo hello world*]
```
hjwp, line 499: i think i'm -1 on this one,  let's introduce the `--rm` flag later if we really want to. for now i want things as simple as possible

chapter_09_docker.asciidoc
```diff
@@ -1417,7 +1420,7 @@ and the `target` path _inside_ the container:
 ----
 $ *docker build -t superlists . && docker run \
   -p 8888:8888 \
-  --mount type=bind,source=./src/db.sqlite3,target=/src/db.sqlite3 \
+  --mount type=bind,source=${PWD}/src/db.sqlite3,target=/src/db.sqlite3 \
```
hjwp, line 1423: does the `.` not work?

hjwp:
thanks so much for these, i really appreciate them!  have asked for a couple of changes or clarifications if poss?

chapter_09_docker.asciidoc
zipperer: The `.` works in current version of docker. I'll revert this proposed edit.

The `.` fails in earlier versions of docker. And I did not try to upgrade docker before I proposed this edit.

## Extra details
### Where the `.` does not work for me
```bash
$ docker --version
Docker version 20.10.14, build a224086
```
and
```
macOS version 11.7.10
```

On that mac I have not found a way to upgrade the docker engine to a more recent version. 
I suspect I won't be able to upgrade the docker engine on that device (see screenshot below).
`${PWD}` works as a replacement for `.` in that setting.

### `.` works in more recent versions of docker
On a separate device where I have
```
macOS Version 12.7.4
```
and 
```
$ docker --version
Docker version 20.10.8, build 3967b7d
```
the `.` failed. But on that machine I was able to upgrade the docker engine
```
$ docker --version
Docker version 26.1.1, build 4cf5afa
```
And after the upgrade the `.` works.

On a third device with Ubuntu and that version of Docker (`26.1.1`), the `.` works.

chapter_09_docker.asciidoc
zipperer: I'll revert the edit to add `--rm`. 

Simple is good. And the reader can clean up later with 
```
docker image prune
docker container prune
```
```
docker rm CONTAINER
docker rmi IMAGE
```
or related commands.

chapter_09_docker.asciidoc
zipperer: Screenshot for users of older version of macOS

20240508-docker-desktop-for-macos-supports-latest-two-version-of-macos

That refers to Docker Desktop and not necessarily the docker engine Docker Desktop uses. But I have not found a way to upgrade the engine independently of Docker Desktop. Savvier users may find a way. 

`${PWD}` instead of `.` is a workaround for users stuck with an old version of docker.

PR #239, opened 2024-05-15
add early return to FT in chapter 13 

Xronophobe:


chapter_13_database_layer_validation.asciidoc
```diff
@@ -808,17 +865,6 @@ We should commit there:
...
 ----
 
-NOTE: So did I break the rule about never refactoring against failing tests?
-    In this case, it's allowed, because the refactor is required
-    to get our new functionality to work.
-    You should definitely never refactor against failing _unit_ tests.
-    But in my book it's OK for the FT for the current story you're working on
-    to be failing.footnote:[
-    If you really want a "clean" test run, you could add a skip or an early return
-    to the current FT, but you'd need to make sure you didn't accidentally forget it.]
```
hjwp, line 818: let's replace this with a `TIP` earlier on in the chapter

```
TIP: When working on a new feature, it's common to want to do a refactor of the application.
  Adding an early return to the FT you're currently working on will let you do that refactor against passing FTs
```

something like that!

We could maybe also add it to the recap at the end?

hjwp:
the tests are failing because of that listing where we moved `item.save()` :

```
[role="sourcecode"]
.src/lists/tests/test_models.py
====
[source,python]
----
    with self.assertRaises(ValidationError):
        item.full_clean()
        item.save()
----
====
//19
// CSANAD: full_clean() should be executed before save() is called. It doesn't
//         make a difference here since the test DB is destroyed anyway, but
// we are actually saving an invalid item and only then calling full_clean().
// Which results in raising the correct error. But I think we should always
// show the correct order even when it doesn't matter: validation first and
// only then attempting save().
```

do you feel comfortable making the fix to the right commit in the chapter 13 branch?

chapter_13_database_layer_validation.asciidoc
```diff
@@ -784,7 +831,17 @@ OK
 ----
 
 
-Let's try a full FT run:
+Let's try a full FT run: they're all passing! We can remove the early return
+now.
```
hjwp, line 835: shall we add a command output showing the fts passing, before we remove the early return?

chapter_13_database_layer_validation.asciidoc
```diff
@@ -808,17 +865,6 @@ We should commit there:
 $ *git commit -am "Refactor list view to handle new item POSTs"*
```
hjwp, line 865: let's move the commit to before we remove the early return, maybe?  wdyt?

chapter_13_database_layer_validation.asciidoc
Xronophobe: I think just telling the reader they are passing should be enough, especially because running them again to find the previous issues in the test is pretty much the same step. Also, the 
```
Ran 4 tests in 20 s

OK
```
output is pretty straightforward. I can add it if you want, but I think it isn't necessary.

chapter_13_database_layer_validation.asciidoc
Xronophobe: Sounds good, how about just right above where we introduce the early return: "We’ve got some working code though, so let’s put an early return in the FT to separate (...)"?

chapter_13_database_layer_validation.asciidoc
```diff
@@ -578,8 +592,47 @@ $ *git commit -am "Adjust new list view to do model validation"*
 ----
 
 
+TIP: When working on a new feature, it's common to realize partway through that
+    a refactor of the application is needed. Adding an early return to the FT
+    you're currently working on allows you to perform this refactor against
+    passing FTs, even while the feature is still in progress.
```
hjwp, line 598: let's move this to the end of the section.  we normally have TIPs _after_ the example of the thing we're suggesting.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -578,8 +592,47 @@ $ *git commit -am "Adjust new list view to do model validation"*
...
+
+
+[role="scratchpad"]
+*****
+* 'Remove hardcoded URLs from views.py'
+* 'Remove the early return from the FT'
+*****
+
+And now, we can focus on making our code a little neater.
+
```
hjwp, line 635: move tip to here i reckon.

chapter_13_database_layer_validation.asciidoc
hjwp: it's not so much for the readers, more for the book's tests - this guarantees that we have passing code/tests at this point in the book.

hjwp:
happy for you to merge when you've addressed the suggestions!  i trust you to make them, don't need to double-check it again...

chapter_13_database_layer_validation.asciidoc
```diff
@@ -566,7 +580,7 @@ Not quite, but they did get a little further.  Checking the line in which the
 error occurred -- 'line 31' in my case -- , we can
 // CSANAD: I think emphasizing what to look for in the error log is helpful.
 see that we've got past the first part of the test, and are now onto the second
-check--that submitting a second empty item also shows an error.
+check -- that submitting a second empty item also shows an error.
```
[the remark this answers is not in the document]
Xronophobe: right, i'm pretty sure I made this mistake before somewhere

chapter_13_database_layer_validation.asciidoc
Xronophobe: in that case, I'll add it of course

PR #242, opened 2024-05-29
Rita's edits to ch 12

ritaf-ORM:


hjwp:
thanks for these rita!  good suggestions as always and much appreciated.

PR #245, opened 2024-05-30
some additions to chapter14

Xronophobe:
add TODO to reorder steps
mention pycache
remove chapter update warning
add commit at the end of the chapter

### DIFF
--- a/chapter_14_simple_form.asciidoc
+++ b/chapter_14_simple_form.asciidoc
@@ -607,7 +607,6 @@ $ *grep -r id_new_item src/lists/*
 
 Good, there are no references to `id_new_item` left--except maybe in some
 `*.pyc` files, which are safe to ignore.
-and they will be regenerated when the code changes.
 What about `name`/ `item_text`?
 
 [subs="specialcharacters,macros"]
--- a/chapter_14_simple_form.asciidoc
+++ b/chapter_14_simple_form.asciidoc
@@ -606,7 +606,7 @@ $ *grep -r id_new_item src/lists/*
 ----
 
 Good, there are no references to `id_new_item` left--except maybe in some
-`*.pyc` files. If you see these, Python generates them for faster execution
+`*.pyc` files, which are safe to ignore.
 and they will be regenerated when the code changes.
 What about `name`/ `item_text`?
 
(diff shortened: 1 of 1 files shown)


chapter_14_simple_form.asciidoc
```diff
@@ -615,9 +605,9 @@ with `id_text` and `text`, respectively:
 $ *grep -r id_new_item src/lists/*
 ----
 
-Good, there are no references to `id_new_item` left.
-// CSANAD: we could mention *.pyc files under __pycache__ would probably still
-//         show up in the results.
+Good, there are no references to `id_new_item` left--except maybe in some
+`*.pyc` files. If you see these, Python generates them for faster execution
```
Xronophobe, line 609: I just figured, pyc files must have been already mentioned somewhere and they were, in Chapter 01 before the first commit. Although, there was no explanation to what they are and why we can ignore them.

hjwp: oh i have `PYTHONDONTWRITEBYTECODE=1` in my env settings, so i just don't see them any more.

not sure i can be bothered ot explain pyc files tbh.

hjwp (approved):

chapter_14_simple_form.asciidoc
Xronophobe: I would mention it at its first occurrence because we are not instructing the readers to set this variable and by default these files would show up. It's just one google search and reading the first hit, but it looks more thorough to save that work maybe in a footnote, in my opinion. 
Ideally, I would move it into a foot note of chapter 01 where we first see pyc files but if you want to, I can just remove it.

hjwp: ok a footnote in chapter 1 pointing people at an explanation on what pycs are sounds good, and then a minimum sentence here?

```suggestion
`*.pyc` files, which are safe to ignore.
```

chapter_14_simple_form.asciidoc
```diff
@@ -615,9 +605,9 @@ with `id_text` and `text`, respectively:
 $ *grep -r id_new_item src/lists/*
 ----
 
-Good, there are no references to `id_new_item` left.
-// CSANAD: we could mention *.pyc files under __pycache__ would probably still
-//         show up in the results.
+Good, there are no references to `id_new_item` left--except maybe in some
+`*.pyc` files, which are safe to ignore.
+and they will be regenerated when the code changes.
```
hjwp, line 610:
```suggestion
```

PR #246, opened 2024-06-06
Update  chapter 15

Xronophobe:
I removed the Queryset Ordering since the problem introducing the topic no longer shows up.
 Suggesting a good article about migration strategies and customizing migration scripts would be great in a footnote, but I haven't found any. I was thinking whether I should write one myself, but I decided not to: probably it's beyond the scope of this chapter (it's called Advanced Forms after all) and deciding what strategy to use would require further discussions anyway.
 Finally, I had to re-add the custom `<input>` to the html template. Conditionally adding classes seemed to be the easiest done this way, and we need that to display the error.


chapter_15_advanced_forms.asciidoc
```diff
@@ -218,216 +210,33 @@ together:
...
-
-((("", startref="triprep15")))((("", startref="queryset15")))That
-works; we get a fully passing test suite:
-
+Great! Our unit tests are passing.
 ----
 OK
 ----
-
+// CSANAD: fully passing suite? I think just the unit tests
```
hjwp, line 229: 👍

chapter_15_advanced_forms.asciidoc
```diff
@@ -218,216 +210,33 @@ together:
...
 
-That long-winded model test did serendipitously help us find an unexpected
-bug, but now it's time to rewrite it. I wrote it in a very verbose style to
-introduce the Django ORM, but in fact, now that we have the explicit test for
-ordering, we can get the same coverage from a couple of much shorter tests.
+That long-winded model test did serendipitously help us find unexpected
+bugs, but now it's time to rewrite it. I wrote it in a very verbose style to
+introduce the Django ORM, but in fact, we can get the same coverage from a
+couple of much shorter tests.
 Delete `test_saving_and_retrieving_items` and replace with this:
```
hjwp, line 240:
```suggestion
Delete `test_saving_and_retrieving_items` and replace it with this:
```

chapter_15_advanced_forms.asciidoc
```diff
@@ -508,10 +317,48 @@ it as an application-layer constraint:
...
+django.db.utils.IntegrityError: UNIQUE constraint failed: lists_item.list_id, lists_item.text
+----
+
+The problem is that we have at least one database record which used to be valid
+but after introducing our new constraint, the `unique_together`, it's no longer
+compatible.
+
+Just delete `src/db.sqlite3` and run the migration again. We can do this,
+because it's only a dev environment, and the data saved in the database isn't
+important.
```
hjwp, line 351:
```suggestion
To fix this problem, we can just delete `src/db.sqlite3` and run the migration again.
We can do this because the database on our laptop is only used for dev, so the data in it is not important.
```

chapter_15_advanced_forms.asciidoc
```diff
@@ -508,10 +317,48 @@ it as an application-layer constraint:
...
+
+However, if we already have records we want to keep (which would be the case in
+a production environment, where we need to keep our users' data) we would need
+to do a bit more work and come up with a migration strategy for dealing with
+invalid--duplicate in this case--records.
+// CSANAD: I wanted to mention some articles from credible sources explaining
+//         migration strategies, modifying migration scripts, etc
+//         but all I found were either outdated, or clearly chatgpt-generated
+// articles.
```
hjwp, line 360: hmmm i think we do mention prod migrations at some point... let me look it up.

hjwp: yes it's mentioned in chapter 17, when we try and deploy the latest code to prod, and there's an appendix on data migrations.  http://www.obeythetestinggoat.com/book/chapter_deploying_validation.html#_what_to_do_if_you_see_a_database_error

hjwp:
```suggestion
In <chapter_deploying_validation>>, we'll deploy our new code to production,
and discuss what to do if we run into migrations and data integrity issues at that point.
```

chapter_15_advanced_forms.asciidoc
```diff
@@ -908,18 +764,78 @@ NOTE: Personal opinion here: I could have used `super`, but I prefer not to use
...
+====
+
+<1> We give `.was-validated` to the `<form>`
+<2> We provide an `<input>` and the most important custom setting will be its
+    `class`. As you can see, we can use conditionals even for providing
+    additional `class` -es.
+<3> If you try to just simply display `form.errors.text` you'll see it ends up
+    as a list item in an unordered list. To avoid that, let's just
+    select `form.errors.text.0`.
+
```
hjwp, line 829:
```suggestion

NOTE:  Another fliip-flop! We spent most of the last chapter switching from handcrafted html
  to having our form autogenerated by django, and now we're switching back.
  It's a little frustrating, and I could have gone back and changed the book's text to avoid the back+forth,
  but I prefer to show software development as it really is.
  We often try things out and end up changing our minds.
  Particularly with frameworks like Django,
  you can find yourself taking advantage of auto-generated shortcuts for as long as they work,
  but at some points you meet the limits of what the framework designers have anticipated,
  and it's time to go back to doing the work yourself.
  It doesn't mean you should always reinvent the wheel!
```

hjwp (approved):

PR #247, opened 2024-06-06
Review Chapter 12

seddonym:
Solid chapter, good reminder that we should apply the same standards of quality to our test code as to our application code.


hjwp:
thanks david!  appreciate the detailed comments :)

PR #248, opened 2024-06-06
Corrected grammatical error. an src -> a src

mshob23:
A small grammatical error I noticed. Love the book!


hjwp:
this is actually intentional, it's how i would say it out loud:  "an ess arr see folder"

totally appreciate it looks weird tho!  😅  not sure what to do about that.

hjwp:
thanks for saying you're enjoying the book!  and thanks also very much for taking the time to submit a patch, i really really appreciate that.  please don't stop!

i'm also very interested in your feedback particularly on these django/deployment/ansible chapters?  they're still very new, so i really want to know about which parts work and which don't....

PR #255, opened 2024-06-08
Re: Rita's review on chapter 12

Xronophobe:


chapter_12_organising_test_files.asciidoc
```diff
@@ -872,10 +862,13 @@ Don't forget the "Refactor" in "Red, Green, Refactor"::
     ((("Red/Green/Refactor")))
 
 Don't refactor against failing tests::
-// RITA: Why shouldn't you refactor against failing tests in general? A brief explanation would be nice here for clarity.
-    * In general!
-// RITA: Please clarify the next line. It's somewhat ambiguous. Are you saying that you should refactor against failing tests in the FT you're currently working on?
-    * But the FT you're currently working on doesn't count.
+    * In general: refactoring against failing tests can easily introduce new
+      issues and make it a lot more difficult to spot them.
+    * Refactoring against failing FTs, however, is often necessary and excusable
```
Xronophobe, line 867: This was a tough one, I spent quite some time phrasing why it's okay to do.
@hjwp do you like/agree with this reasoning?

hjwp: thanks!   i appreciate the effort.  i think maybe less is more tho... i want to stay away from the "what vs how" distinction because i think it's good even in unit tests to keep the focus on behaviour ie what rather than how.

so how about we keep it simpler, and say something like

> The general rule is that you shouldn't mix refactoring and behaviour change.  Having green tests is our best guarantee that we aren't changing behaviour.  If you start refactoring against failing tests, it becomes much harder to spot when you're accidentally introducing a regression.

> This applies strongly to unit tests.  WIth functional tests, because we often develop against red FTs anyway, it's sometimes more tempting to refactor against failing tests.   My suggestion is to avoid that temptation and use an early return, so that it's 100% clear if, during a refactory, you accidentally introduce a regression that's picked up in your FTs.

Issue #240, opened 2024-05-15
Broken code listing id-s in Chapter 13


Xronophobe:
The code listings' id-s in Chapter 13 refer to Chapter 11, e.g.: `(ch11l018)`
When I added another, I chose to just write `(ch13l???)`.

hjwp:
yes.  it's a minor thing, but if we have time, it might be nice to rewrite all the commit messages for this chapter, and fix all those listings...

PR #257, opened 2024-06-12
Fixes for tests in 13

hjwp:
The tests here were failing because a couple of changes we made to the listings were not in sync witht he commit history of the branch in the `book-example` repo.  Here's how I fixed things at a high level:

* first, determine that the cause was due to the commits getting out of sync.  I saw the error was that the Fts were failing when the book said they should be passing, and then I see it's at a part of the book where we say "now remove the early return", and I realised we didn't have a commit where we remove the early return.
* I then realised we didn't have a commit for when we *added* the early return either!  that's where we had the commit listing as `ch13l???`.
* so I first added a commit for when we introduce the early return.  I did this by:
1. checkout the chatper 13 branch
2. `reset --hard` to the commit just _before_ where we want to add the early return code
3. add the early return, and make a new commit for it.  I put `--ch13l030-3--` at the end of the commit message
4. update the book to show this "commit id" (i also tweaked the comment message while i was at it)
5. `git cherry-pick` all the rest of the commits from the old commit history

that procedure, `reset --hard <commit N>` + `commit <new commit N+1>` + `git cherry-pick <old-commit-N+1...end>` is one i use quite often.  it's a bit like a rebase, rewriting history.

Then I followed a similar procedure to introduce an extra commit for when we _remove_ the early return as well.  I added that to the book chapter text too.

Finally I noticed a problem with applying one of the later commits, it couldn't apply the commit patch successfully.  I managed to figure out that was because of the change where we swapped the order of `form.save()` and `form.full_clean()`.  SO for that I did a `git rebase --interactive` and marked the commit where we introduce that form.full_clean() for editing, and changed it to match the book.

that gave me a conflict at the end of the rebase, which is expected, and so i fixed that..... i'm aware this explanation is getting a little abstract probably.  but i thought it was good to write it down.

happy to go over this sort of procedure in detail with you one of these days!  there just never seems to be enough timeeeeeeee.  but, you have to invest time to make time! so.

### DIFF
--- a/chapter_13_database_layer_validation.asciidoc
+++ b/chapter_13_database_layer_validation.asciidoc
@@ -112,14 +112,6 @@ class ListAndItemModelsTest(TestCase):
 ----
 ====
 
-TIP: If you're new to Python, you may never have seen the `with` statement.
-    It's used with what are called "context managers", which wrap a block of code,
-    usually with some kind of setup, cleanup, or error-handling code.
-    There's a good write-up in the
-    https://docs.python.org/release/2.5/whatsnew/pep-343.html[Python 2.5 release notes].
-    ((("with statements")))
-    ((("Python 3", "with statements")))
-
 This is a new unit testing technique: when we want to check that doing
 something will raise an error, we can use the `self.assertRaises` context
 manager.  We could have used something like this instead:
@@ -143,6 +135,15 @@ and see its expected failure:
     with self.assertRaises(ValidationError):
 AssertionError: ValidationError not raised
 ----
+
+
+TIP: If you're new to Python, you may never have seen the `with` statement.
+    It's used with what are called "context managers", which wrap a block of code,
+    usually with some kind of setup, cleanup, or error-handling code.
+    There's a good write-up in the
+    https://docs.python.org/release/2.5/whatsnew/pep-343.html[Python 2.5 release notes].
+    ((("with statements")))
+    ((("Python 3", "with statements")))
 
 
 A Django Quirk: Model Save Doesn't Run Validation
@@ -598,7 +599,7 @@ Let's put an early return in the FT to separate
 what we got working from those that still need to be dealt with:
 
 [role="sourcecode"]
-.src/functional_tests/test_list_item_validation.py (ch13l???)
+.src/functional_tests/test_list_item_validation.py (ch13l030-3)
 ====
 [source,python]
 ----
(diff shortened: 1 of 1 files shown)


chapter_13_database_layer_validation.asciidoc
```diff
@@ -598,7 +599,7 @@ Let's put an early return in the FT to separate
 what we got working from those that still need to be dealt with:
 
 [role="sourcecode"]
-.src/functional_tests/test_list_item_validation.py (ch13l???)
+.src/functional_tests/test_list_item_validation.py (ch13l030-3)
```
hjwp, line 602: new commit id, see new history in https://github.com/hjwp/book-example/commits/chapter_13_database_layer_validation/

chapter_13_database_layer_validation.asciidoc
```diff
@@ -608,8 +609,7 @@ class ItemValidationTest(FunctionalTest):
         self.browser.find_element(By.ID, "id_new_item").send_keys(Keys.ENTER)
         self.wait_for_row_in_list_table("1: Buy milk")
 
-        return
-        # TODO fix the remaining test scenarios
+        return  # TODO re-enable the rest of this test.
```
hjwp, line 612: decided to tweak this listing slightly. sorry i'm a micromanager!  😅

chapter_13_database_layer_validation.asciidoc
```diff
@@ -695,13 +695,13 @@ This will immediately break our original functional test, because the
 ----
 $ pass:quotes[*python src/manage.py test functional_tests*]
 [...]
-selenium.common.exceptions.NoSuchElementException: Message: Unable to locate
-element: .invalid-feedback; [...]
-[...]
 AssertionError: '2: Use peacock feathers to make a fly' not found in ['1: Buy
```
hjwp, line 698: a nice side-effect of this fix - the book tests failed on this bit, saying it couldn't find this part of the traceback.  sure enough, it's because this error was coming from the FT that we've put an early return in, so it doesn't actually appear any more.

this is exactly what the book tests are for, yay!

chapter_13_database_layer_validation.asciidoc
Xronophobe: This is related to https://github.com/hjwp/Book-TDD-Web-Dev-Python/issues/240

chapter_13_database_layer_validation.asciidoc
Xronophobe: Oh that's a nice catch!

PR #259, opened 2024-06-16
Fix a listing in 14

hjwp:
hopefully will get the tests passing.

* the failure was due to the fact that the test runner wraps console output at 80 chars.
* that hid a second failure, which is that when we report `NoSuchElementException`, selenium prints a whole load of carp, that we basically have to hide in the output listings with a trailing `[...]` on the line.


Xronophobe:
awesome!!! thanks!

PR #274, opened 2024-07-02
Update chapter_12_organising_test_files.asciidoc: fix link

gasull:
Fix link to the chapter The Token Social Bit, the Page Pattern, and an Exercise for the Reader.

The link was broken when reading directly on GitHub.


hjwp:
thanks for the contribution!

i'm afraid this might work in the rendered versions on github, but github isn't really a supported / intended publication target; the link you propose won't work for the online (or pdf) versions of the book.  see eg http://www.obeythetestinggoat.com/book/chapter_12_organising_test_files.html where you'll see current anchor works fine.

if you're keen to try this out locally, you can clone the repo and do `make book.html` to test out a local html render of the whole book, which lets you check that inter-chapter links work...

PR #283, opened 2024-07-25
Review by David Seddon of database layer validation

seddonym:
Good chapter, mostly small comments. But I do wonder if framing this as 'testing at the database layer' is a good description of what we're doing. Database layer validation would involve running SQL I think.

### DIFF
--- a/chapter_13_database_layer_validation.asciidoc
+++ b/chapter_13_database_layer_validation.asciidoc
@@ -102,6 +102,9 @@ class ListAndItemModelsTest(TestCase):
 ----
 ====
 
+// DAVID: Catching ValidationError is a bit too much of a catch-all, could
+// we assert what's on the ValidationError for the correct error?
+
 This is a new unit testing technique: when we want to check that doing
 something will raise an error, we can use the `self.assertRaises` context
 manager.  We could have used something like this instead:
@@ -136,6 +139,10 @@ TIP: If you're new to Python, you may never have seen the `with` statement.
     ((("Python 3", "with statements")))
 
 
+// DAVID: Surely there's a better link than something from Python 2.5,
+// which contains Python examples that aren't compatible with Python 3?
+// Some readers may not be aware of the differences and could trip them up.
+
 A Django Quirk: Model Save Doesn't Run Validation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


chapter_13_database_layer_validation.asciidoc
```diff
@@ -201,9 +213,12 @@ Let's hack it in to see it work:
 // Which results in raising the correct error. But I think we should always
 // show the correct order even when it doesn't matter: validation first and
 // only then attempting save().
+// DAVID: A bit like Csanad, why bother calling item.save() at all?
 
 That gets the unit test to pass:
 
+// DAVID: Would be good to provide the command to run the unit test.
```
hjwp, line 220: i actually gave up on reminding people how to run the unit tests back in chapter 5 or so.  we do it an awful lot!

chapter_13_database_layer_validation.asciidoc
```diff
@@ -521,6 +536,7 @@ class NewListTest(TestCase):
         self.assertEqual(Item.objects.count(), 0)
 ----
 ====
+// DAVID: Suggest using self.assertFalse(List.objects.exists()), etc.
```
hjwp, line 539: which assertion gives the most useful error message when it fails?  i'd have thought `assertEqual(...objects.all(), [])` actually....

chapter_13_database_layer_validation.asciidoc
```diff
@@ -760,7 +778,10 @@ class ListViewTest(TestCase):
 Note that the `NewItemTest` class disappears completely.
 I've also changed the name of the redirect test to make it explicit that it only applies to POST
 requests.
+// DAVID did you? Looks like the same name to me?
+
 
+// DAVID: we switched to running the unit test, would be good to provide the command.
```
hjwp, line 784: are you still going through and doing all the listings on your own pc?  it should be obvious, we've just edited a unit test file.  i expect readers not to need too much hand-holding at this point in the book...

chapter_13_database_layer_validation.asciidoc
```diff
@@ -1252,6 +1284,9 @@ That final scratchpad item will be the subject of the next chapter.
 I always like to push my validation logic down as low as possible.
 ((("model-layer validation", "benefits and drawbacks of")))
 
+// DAVID: Are you, though? This is only validation at the Python layer, not the database layer.
+// Better to test against some SQL that tries to insert it?
```
hjwp, line 1288: this is why we had the discussion of `.save()` raising `IntegrityError` straight from the db, vs `.full_clean()` raising `ValidationError` without actually touching it.  let me see if i can clarify that section...

hjwp: aaaah ok i see, re-reading the chapter text now, we never actually managed to get an IntegrityError out.  We do in the next chapter!  will rephrase, thanks David.

PR #286, opened 2024-07-28
Review of chapter 13

Enforcer:
It is a rather quick and pretty enjoyable chapter.

There is not much of a constructive feedback from me.

One more serious remark I have is that I got lost when we jumped from Model-level validation to `assertRaises`. I suggest adding some linking sentences.


hjwp:
thanks Sebastian, appreciate all three suggestions!

PR #297, opened 2024-08-12
Review of chapter 14

Enforcer:
This is an enjoyable chapter that's half about Django's [Model]Form and half about testing.

I had fun reading it. I found very little things to complain about.

One bigger thing is that I'd put at the beginning something to set the expectations and do a little priming of the reader - that we're indeed gonna do be doing some refactor and in such a case functional tests shield us.

Even though we had to make some changes, it was only adjusting details here and there. Behaviour did not change even though we refactored significant amount of code and logic to leverage ModelForm

### DIFF
--- a/chapter_14_simple_form.asciidoc
+++ b/chapter_14_simple_form.asciidoc
@@ -191,6 +191,9 @@ Here we're actually using a unit test as a way of experimenting with the forms A
 It's actually a pretty good way of learning how it works.
 *******************************************************************************
 
+// SEBASTIAN: Small suggestion - I'd appreciate mentioning breakpoint() for use in test
+//      to be able to play with a form instance even more
+
 
 ==== Switching to a Django ModelForm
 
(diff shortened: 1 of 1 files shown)


chapter_14_simple_form.asciidoc
```diff
@@ -536,13 +542,15 @@ A quick `grep` shows us there are several places where we're using `id_new_item`
...
 self.browser.find_element(By.ID, "id_new_item")
-src/functional_tests/test_layout_and_styling.py:        inputbox =
+src/functional_tests/test_layout_and_styling.py:        inputbo§§x =
 self.browser.find_element(By.ID, "id_new_item")
 src/functional_tests/test_list_item_validation.py:
 self.browser.find_element(By.ID, "id_new_item").send_keys(Keys.ENTER)
 [...]
 ----
 
+// SEBASTIAN: Are those "§§" not a mistake here? 3rd line of grep result
```
Xronophobe, line 552: They definitely look like a mistake, but your commit introduced them. I was wondering if it was accidental until I saw your comment. I'm removing this change,

PR #298, opened 2024-08-12
Review of chapter 15

Enforcer:
This chapter felt much lighter than the previous one but also nicely extends the ideas presented there. It's a good choice to keep them separate.

I fixed two typos and added a bunch of rather small comments.

One more serious thing from my perspective is that the summary of testing methods is a bit confusing at the moment and the lack of blank lines made my brain melt at the first encounter. Can something be done to make it less overwhelming?

At the end, I'm bit puzzled tho - do we retain ItemForm in the end or shall we replace it fully with a new custom Form class?
Why do we still need/want the old one?


hjwp:
thanks Sebastian!

chapter_15_advanced_forms.asciidoc
```diff
@@ -141,6 +141,17 @@ it can be hard to see what's being tested:
...
+// SEBASTIAN: for scenarios that are meant to test if something IS NOT raising exception, I use:
+// ```
+// def test_():
+//     ...
+//     try:
+//         model.full_clean()
+//     except ValidationError:
+//         self.fail("Should not raise exception!")
+// ```
+// SEBASTIAN continued: WDYT?
```
hjwp, line 153: @Enforcer nice!  you'd. get a more explicit error message in the test fail this way.  otoh it is a bit more verbose...

PR #299, opened 2024-08-13
Review of chapter 17

Enforcer:
There is not much content in this chapter which actually... is a good thing here.

It shows one specific, yet crucial thing - [semi-manual] deployment pipeline with testing in staging before deployment to production.

It's great that potential issues with migrations are mentioned and some solutions for staging are shown.

I was a bit surprised by tagging the release part. So far I understood it differently - to put a tag onto a new about-to-be-released version of the application, not specifically tagging what was deployed when and what is currently live. Although, the thing that is presented makes sense.


hjwp:
thanks buddy! :)

PR #301, opened 2024-08-13
wip address tech review comments

Xronophobe:


chapter_12_organising_test_files.asciidoc
```diff
@@ -466,7 +467,6 @@ might decide that building a specific helper method is overkill at this stage,
 but it might be nice to have some generic way of saying, in our tests, "wait
 until this assertion passes".  Something like this:
 
-// DAVID: might be worth reminding the reader about that helper method, e.g. what it's called.
```
Xronophobe, line 469: in my opinion it isn't the focus here (I mean we just merely mention this other method)

chapter_12_organising_test_files.asciidoc
```diff
@@ -517,8 +517,6 @@ and we'll adapt it slightly:
 ----
 ====
 
-// DAVID: I have `raise e` from that other method, not `raise`.
```
Xronophobe, line 520: I checked in Chapter 6 and it should be just `raise`. We hadn't dealt with the raised error object earlier and after having quickly checked I believe the first time we will is in Chapter 20.

chapter_12_organising_test_files.asciidoc
```diff
@@ -602,8 +600,6 @@ and that can be executed later, and multiple times:
 
 Let's see our funky `wait_for` helper in action:
 
-// DAVID: Might be worth explicitly telling the reader to add that code earlier - I hadn't yet,
```
Xronophobe, line 605: I think the previous few paragraphs (ch11l010) were clear about implementing this. Has anyone else find it unclear though? If so, we can try making it clearer.

chapter_12_organising_test_files.asciidoc
```diff
@@ -652,12 +648,6 @@ through what happened:
     and inside the body of the `lambda` function,
     and tells us that it was attempt to find the `.invalid-feedback` element
     that failed.
-// CSANAD: I changed two things here:
```
Xronophobe, line 655: I added these comments along with my changes while both reviewing and updating chapter 12. Since these changes have already been merged, the explanations are no longer needed.

chapter_12_organising_test_files.asciidoc
```diff
@@ -71,9 +70,8 @@ how to get there step by step.
 ==== Skipping a Test
 
 NOTE: We're back to local development now.
-    Double check that the `TEST_SERVER` environment variable is unset in your terminal.
-
-// DAVID: might be nice to remind people how they check this?
```
Xronophobe, line 76: agreed, added

Xronophobe, line 76: I agree, added

chapter_12_organising_test_files.asciidoc
```diff
@@ -385,8 +378,8 @@ Brilliant--no need to sit around waiting for all the FTs
 when we're only interested in a single one.
 Although we need to remember to run all of them now and again, to check for regressions.
 Later in the book we'll set up a Continuous Integration (CI) server to run all the tests automatically,
-for example every time we push to master.
-// DAVID: "master" -> "our main branch"
+for example every time we push to the default branch (which is the `main` or
+formerly known as the `master` branch).
```
Xronophobe, line 382: I would keep a reference to `master`. As I checked (and found a URL pointing to the `master` tree in chapter 02) we refrained from naming the default branch so far. Also, the official git documentation still uses the original name for the default branch. So I think it's best to just call it default, and maybe mentioning the two most common names for it.

chapter_12_organising_test_files.asciidoc
```diff
@@ -874,6 +855,7 @@ Use a tests folder::
       'test_forms.py'.
 // DAVID: worth mentioning that mirroring test files to real files can pave the way for automatically running tests
 // relevant to your changes.
+// CSANAD: TODO ask David to elaborate
```
Xronophobe, line 858: @seddonym Could you please explain what you mean by this specifically? Or maybe would you like to try phrasing it yourself?

chapter_12_organising_test_files.asciidoc
```diff
@@ -205,7 +199,6 @@ class ItemValidationTest(FunctionalTest):
 ----
 ====
 
-// DAVID: I also have check_for_row_in_list_table as a method - maybe I should've removed it?
```
Xronophobe, line 208: Yes, it was in Chapter 06, right above "Example 4. functional_tests/tests.py (ch06l004)". We renamed this function.

Xronophobe:
Hi @seddonym thanks for the review! Nice catches.

chapter_12_organising_test_files.asciidoc
seddonym: We have git hooks that automatically run test files based on the changed files. So for example if you change a file called `path/to/somemodule.py` and the tests are located in (say) `tests/path/to/test_somemodule.py` then we can automatically run those tests.

@hjwp will be familiar with this as it's something we do at our workplace.

hjwp: indeed!  i'm not sure i'm really a fan of the whole idea tho.

chapter_12_organising_test_files.asciidoc
hjwp: it ends up quite wordy tho.  i quite like david's phrasing, which neatly leaves it ambiguous (if your default branch is called master, it's still your main branch, even if it's not literally named main).  i don't think people will be confused.

chapter_12_organising_test_files.asciidoc
```diff
@@ -168,9 +165,6 @@ https://www.oreilly.com/library/view/tidy-first/9781098151232/[Tidy First?]
 ((("test files", "splitting FTs into many")))
 We start putting each test into its own class, still in the same file:
 
-// DAVID: I would say this is too big a refactor all in one go.
```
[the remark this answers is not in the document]
hjwp: i don't have strong feelings about it really -- do you think david's idea of how to do the refactor is "smaller"?  i kinda think they're equivalent really.   not sure it's really worth making the change.  but i'm willing to be convinced!

PR #302, opened 2024-08-14
Csanád address ch 13 review comments

CsanaadB:
TODO:

- [x]  "Well, for
http://bit.ly/2v3SfRq[slightly counterintuitive historical reasons],
Django models don't run full validation on save.
// CSANAD: I'd argue against the use of shortened URLs."
=> https://groups.google.com/g/django-developers/c/uIhzSwWHj4c ? 
Check URL first

### DIFF
--- a/chapter_13_database_layer_validation.asciidoc
+++ b/chapter_13_database_layer_validation.asciidoc
@@ -147,7 +147,8 @@ Which means, since TextField is a type of Field, it _should_ disallow empty valu
 ((("data integrity errors")))
 So why is the test still failing?
 Well, for
-http://bit.ly/2v3SfRq[slightly counterintuitive historical reasons],
+https://groups.google.com/g/django-developers/c/uIhzSwWHj4c[slightly
+counterintuitive historical reasons],
 Django models don't run full validation on save.
 As we'll see later,
 any constraints that are actually implemented in the database will raise errors on save,


chapter_13_database_layer_validation.asciidoc
```diff
@@ -75,10 +72,6 @@ use to validate input, but you're always going to use the same database.
 
 ==== The self.assertRaises Context Manager
 
-// SEBASTIAN: This seems to come out of the blue, at least a bit.
```
Xronophobe, line 78: At first glance I agreed, but then I re-read the chapter from the beginning and it feels it's been linked properly by the second paragraph above.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -106,8 +99,6 @@ class ListAndItemModelsTest(TestCase):
 ----
 ====
 
-// DAVID: Catching ValidationError is a bit too much of a catch-all, could
-// we assert what's on the ValidationError for the correct error?
```
Xronophobe, line 110: I think this would be more valid had we expected raising just a generic Exception. In this test case, in my opinion, ValidationError is a sufficiently specific Exception, given `mylist` does have a valid value.
Also, there aren't too many more specific things to assert, in my opinion. We will provide a "You can't have an empty list item" error message but that will be `views`' territory.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -173,6 +164,9 @@ so if we try to run `makemigrations`, it will report there's nothing to do:
 // CSANAD: maybe I'm missing something here but I don't think there would have
 //         been any changes detected here even with another DB. I don't remember
 // changing the model since the last migration.
+// Update: Checked and last time we touched models.py was in chapter 07 where we did
+// run makemigrations and then `migrate` in chapter 08. No changes to the model since.
+// Can I delete this part "There's a way of checking whether the constraint..."?
```
Xronophobe, line 169: fyi @hjwp

chapter_13_database_layer_validation.asciidoc
```diff
@@ -262,7 +249,6 @@ been passed an error variable, and if so, we do this:
 <1> We add the `.is-invalid` class to an form inputs with validation errors
 <2> We use a `div.invalid-feedback` to display any error messages from the server.
 
-// RITA: Perhaps put the Bootstrap docs note and the footnote text into a tip box?
```
Xronophobe, line 265: forgot to remove earlier, but already done

chapter_13_database_layer_validation.asciidoc
```diff
@@ -325,8 +303,6 @@ and make a note to refactor them out.  But we won't do them straight away,
 because right now our application is in a broken state. We want to get back
 to a working state first.
 
-// JAN: It's preferred to use hardcoded URLs in tests. If we use resolving functions it's easy to change URLs and have passing tests while actually breaking stuff. For example, all the previously sent emails that the old URL.
```
Xronophobe, line 328: I think for this problem is best to have one test to check whether the result of calling `reverse()` is the same as the hardcoded url. And not to use hardcoded URLs anywhere else.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -859,7 +836,6 @@ urlpatterns = [
 ----
 ====
 
-// JAN: I have 3 URL patterns - incl. path("", views.home_page, name="home"),
```
Xronophobe, line 862: that should be in `src/superlists/urls.py` but this is `src/lists/urls.py`

chapter_13_database_layer_validation.asciidoc
```diff
@@ -876,7 +852,6 @@ Ran 4 tests in 9.951s
 OK
 ----
 
-// DAVID: I get four "Not Found: /favicon.ico" messages when running the FTs, just FYI.
```
Xronophobe, line 879: me too, but I think it's no big deal. I think the HTML is referencing it by default.

chapter_13_database_layer_validation.asciidoc
```diff
@@ -1317,8 +1292,6 @@ And it's not designed for user-friendliness::
     Trying to store invalid data will cause a nasty `IntegrityError` to come
     back from your database, and possibly the user will see a confusing 500
     error page.
-    // DAVID: We haven't seen any IntegrityErrors here? This feels like it doesn't
-    // apply to this chapter.
```
Xronophobe, line 1321: We have seen a few in earlier chapters but it's true, we have not triggered any in this one. But the statement is true though, attempting to store invalid data does result in IntegrityErrors. I would keep it.

Xronophobe:
Thank you @jangia and @seddonym for the review! Code changes are a little complicated so I just left those review comments and marked where I'd agree with the change (and working on them). For the rest, I added comments. Feel free to argue with my responses if you disagree!  Happy to discuss them.

chapter_13_database_layer_validation.asciidoc
hjwp: i have some changes i want to tmake tot this chapter in general, maybe deliberately raising an actual IntegrityError.  ok to delete... or maybe comment out this section?  with a `////` ?

chapter_13_database_layer_validation.asciidoc
```diff
@@ -782,6 +761,8 @@ class ListViewTest(TestCase):
 //      looks like something that could be changed to a declarative sentence without noise like "can"
 //      maybe like "POST_on_existing_list_adds_item_to_the_list"?
 //      Also, from what I can see, other_list is not used at all in both tests. Is this intended?
+// CSANAD: not sure how it slipped my attention, true. We don't need the `other_list` in these
+// test cases and I might as well rename the test case
```
hjwp, line 765: i like explanatory variable names.  and i can't be bothered with the faff of feeding thru the commits for a test name change lol

hjwp (approved):

hjwp (approved):

PR #303, opened 2024-08-17
Add comments on chapter 9

mmxw:
Thanks for a good read, Harry!
Added some in-line comments for chapter 9.

### DIFF
--- a/chapter_09_docker.asciidoc
+++ b/chapter_09_docker.asciidoc
@@ -72,6 +72,7 @@ I hope you enjoy the new version!
 ****
 
 === TDD and the Danger Areas of Deployment
+// XUAN: I read the title as meaning _TDD and_ the danger areas of deployment, but the following section seems to be talking about the danger areas of deployment?
 
 Deploying a site to a live web server can be a tricky topic.
 Oft-heard is the forlorn cry __"but it works on my machine!"__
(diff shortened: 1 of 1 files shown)


chapter_09_docker.asciidoc
```diff
@@ -72,6 +72,7 @@ I hope you enjoy the new version!
 ****
 
 === TDD and the Danger Areas of Deployment
+// XUAN: I read the title as meaning _TDD and_ the danger areas of deployment, but the following section seems to be talking about the danger areas of deployment?
```
hjwp, line 75: thanks @mmxw !  yes i think this dates back to a previous edition where i linked more explicitly between the danger areas and how tdd can help.  now i think we're more talking about how docker can help, by letting us take small steps and test things locally... with tdd still, but kind of later in the loop.

will update the title to remove the tdd bit probably.

chapter_09_docker.asciidoc
```diff
@@ -276,6 +278,8 @@ How will containerizing our software help with the danger areas?
 
 * Later, when we deploy our containers to a staging server,
   we can run the FTs against that too.
+// XUAN: "FTs" = "functional tests"?
+// XUAN: This is the first time this chapter mentions "staging", which seems a bit out of the blue. Can you elaborate on what problems does having a staging server solve for the initial list of danger areas of deployment?
```
hjwp, line 282: thanks, yes!  i think maybe a more explicit discussion of the incremental approach, docker first, run tests against that, staging second, run tests against that, prod third...

PR #308, opened 2024-08-21
Csanad: review chapter 17

Xronophobe:
Quick, fast chapter but I found a bug which I mentioned to Harry in a message. We had a discussion whether we should update chapter 11 or include the (short, one-line) solution here.

I'm personally on the opinion to extend the `ansible-playbook.yaml` here in chapter 17, when we notice the bug after running the FTs against the staging server. To me, it feels more natural not to include `force_source` in the "Import container image on server" task until we try to create a second deployment here, in chapter 17 where we notice that somehow the deployed version is old.
As Harry pointed out, some readers may try to deploy their code (therefore also face this bug) sooner. Still, I think we could include the solution in this chapter but mention it in a NOTE in chapter 11.


hjwp:
thanks buddy!

PR #316, opened 2024-09-04
Better discussion of IntegrityError vs validation errors in 13

hjwp:


chapter_13_database_layer_validation.asciidoc
```diff
@@ -81,155 +81,270 @@ This test is interesting
...
 
 [role="sourcecode"]
-.src/lists/tests/test_models.py
+.src/lists/tests/test_models.py (ch13l004)
 ====
 [source,python]
 ----
     with self.assertRaises(ValidationError):
         item.full_clean()
-        item.save()
```
Xronophobe, line 204: are we removing the `save()` and changing it to a `full_clean()` or are we just inserting the `full_clean` in front of `save()`?

hjwp: changing it.  i'll add a note!

PR #332, opened 2024-12-10
Csanád's review of chapter 19

Xronophobe:
I like the email-based authentication, even though I had mixed feelings about it initially (not in the book specifically, but about Magic Links in general).

In the "spike" part of the chapter we sometimes take bigger leaps without checking our work, which I found a bit difficult to keep up with at first. Although we can't always check our work step-by-step, and having to re-read a few paragraphs here and there is just how technical books work. Also, practicing to read and make sense of source code is how developers operate. So I would not necessarily change it, I'm just providing my impressions. More importantly, I found it extremely useful: showing the readers the thought process, the sources/docs used for discovering how we could make something (like this custom authentication) work.

A couple of further notes without checking whether later chapters address these:
- not sure if we'll delete the tokens on `/logout`. I guess we certainly will, but I'm leaving a note regardless.
- either way, I would still hash the token in the database, similar to how `set_password()` works in a password-based Django auth model. After all, a token alone grants full access to the user's account, so I would handle it the way I would handle a password.

### DIFF
--- a/chapter_19_spiking_custom_auth.asciidoc
+++ b/chapter_19_spiking_custom_auth.asciidoc
@@ -91,6 +91,7 @@ but this is just a fun toy project so let's give it a go.
 To get this Magic Links project set up, the first thing I did was take a look at existing Python and Django authentication
 packages, like https://docs.allauth.org/en/latest/[django-allauth]
 and https://github.com/omab/python-social-auth[python-social-auth],
+// CSANAD: python-social-auth looks abandoned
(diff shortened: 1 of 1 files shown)


hjwp (approved):

hjwp:
thanks csanad!  hash the tokens, yep, pretty obvious in retrospect. this is why you don't roll your own security folks!     i might actually leave it as-is but call it out explicitly in a sidebar...

PR #333, opened 2024-12-19
Review of simple form chapter

seddonym:
Good stuff. I like the way you gradually introduce the idea of a form here.

Reviewing this chapter prompted a couple of more general thoughts.

# Functional tests

The Selenium-based functional testing approach involves tradeoffs. For me, the lower-level Django test client (or, even better [django-webtest](https://pypi.org/project/django-webtest/)) is a sweeter spot for functional testing. I appreciate we use these in test_views, but they're not really unit tests. I'm not sure whether you get to it later in the book, but I think it would be worthwhile discussing the different options so people don't feel like they absolutely have to use Selenium to have decently tested web applications.

And, more generally - we can't test _everything_, right? So it's about judging the tradeoffs. I think the readers should feel empowered to see that this is just one way but they should keep reflecting on whether it's a way that works for them. My tests for Django look very different to the ones in this book - both the unit and the functional tests. 

# Keeping the test suite passing

One of the best things I've taken from our recent TDD dojos is the idea of keeping the test suite passing during refactors, even to the point of doing apparently counter-intuitive things such as copy-pasting code. In this chapter we have the test suite failing for quite a while (even to the point where we're making git commits while the suite is failing).

# `+` in tips at end

Re. the stray `+`, here's a screenshot of how it appears to me locally. Maybe it's a problem with how I'm building it though.

image

### DIFF
--- a/chapter_15_simple_form.asciidoc
+++ b/chapter_15_simple_form.asciidoc
@@ -1205,6 +1205,10 @@ using a form is a nice way of encapsulating that logic.
 // for this kind of thing, which doesn't have browser validation built in.
 // So we could use that if we really wanted to test the server side validation.
 //
+// [Update]: I just clocked that we are doing exactly this in test_views.py. Might be worth just calling it out a bit more?
+//
+
+//
 // This might be a good opportunity to mention that as an option
 // (even if we don't go into detail here)
 // and discuss the trade offs of using Selenium versus the test client, versus both.
--- a/chapter_15_simple_form.asciidoc
+++ b/chapter_15_simple_form.asciidoc
@@ -119,7 +119,8 @@ class ItemFormTest(TestCase):
         self.assertIn('class="form-control form-control-lg"', form.as_p())
 ----
 ====
-
+// DAVID: what about splitting this up into arrange/act/assert and calling .as_p() only once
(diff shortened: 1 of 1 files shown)


chapter_15_simple_form.asciidoc
```diff
@@ -1183,6 +1198,16 @@ so you always need the server side as well
 if you really care about data integrity;
 using a form is a nice way of encapsulating that logic.
 
+// DAVID: I wonder if this is slightly missing the point.
+// Our functional tests here are very high level - using a headless browser.
+// But many developers use the Django test client
+// (https://docs.djangoproject.com/en/5.1/topics/testing/tools/#the-test-client)
+// for this kind of thing, which doesn't have browser validation built in.
+// So we could use that if we really wanted to test the server side validation.
```
seddonym, line 1206: Update: I just clocked that we are doing exactly this in `test_views.py`. Might be worth just calling it out a bit more?

hjwp:
```suggestion
// So we could use that if we really wanted to test the server side validation.
//
// [Update]: I just clocked that we are doing exactly this in test_views.py. Might be worth just calling it out a bit more?
//

```

PR #334, opened 2024-12-19
Review of advanced forms chapter

seddonym:
Looking good. A few small issues that I noticed though.

A general point: I feel like just because we're doing TDD doesn't mean we can't occasionally start up the application and use it to figure out what's happening. It feels like a long time since we've done that!


chapter_16_advanced_forms.asciidoc
```diff
@@ -385,6 +400,8 @@ We can do this because the database on our laptop is only used for dev, so the d
 In <<chapter_18_second_deploy>>, we'll deploy our new code to production,
 and discuss what to do if we run into migrations and data integrity issues at that point.
 
+// DAVID: Shall we get them to run migrations again, since they deleted the database?
```
seddonym, line 403: Update - we definitely should do this because otherwise when they try to start up the application in a subsequent chapter, it won't work.

PR #339, opened 2024-12-20
Review of chapter 19

Enforcer:
This chapter is pure perfection 👌

It reads nice. There wasn't a single time I felt confused or overwhelmed - the cognitive load was just right.

Also, I learnt something.

My only comment was to question whether a paragraph about the history of this chapter is of any value to the user.


hjwp:
aw Sebastian what a lovely thing to say thank you so much!

yep i think i'll move that history to a footnote or something.

PR #345, opened 2024-12-20
Review of mocking 1

seddonym:
I like the pragmatic approach you take to mocking here (especially consider we both know they are BAD!) In particular I think the messages framework is a great way of making the point.

I think a bit more explanation for the Django bits wouldn't go amiss - and as I say in the comment, perhaps even ditch the whole authentication backend for a lower-level approach that uses the session framework directly.

It might also be worth drawing the distinction between mocking and monkeypatching, and the trade offs involved with each. Maybe also some more strongly worded advice against directly monkeypatching attributes in tests without tidying up afterwards.

### DIFF
--- a/chapter_20_mocking_1.asciidoc
+++ b/chapter_20_mocking_1.asciidoc
@@ -27,6 +27,8 @@ so for the purpose of these unit tests, we'll pretend that this nice Django shor
 
 Am I telling you _not_ to use Django's `mail.outbox`?
 No; use it, it's a neat shortcut.
+// DAVID: More than just a shortcut, it's a much better approach than mocking because it's
+// a documented, swappable API.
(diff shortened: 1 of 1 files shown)


chapter_20_mocking_1.asciidoc
```diff
@@ -1429,5 +1465,5 @@ Mocks can leave you tightly coupled to the implementation::
   As we saw in <<mocks-tightly-coupled-sidebar>>,
   mocks can leave you tightly coupled to your implementation.
   For that reason, you shouldn't use them unless you have a good reason.
-
+// DAVID: When rendered, this last point says the same phrase three times in a row.
```
seddonym, line 1468: image

hjwp: 😂 thanks for that

PR #346, opened 2024-12-20
Review of chapter 21

Enforcer:
It is a pretty long chapter, full of valuable knowledge and some key considerations, e.g. regarding a number of tests, preventing combinatorial explosion and mocks in general.

Although it reads rather easily, it feels long. While reading the previous chapter, I didn't have that impression.

I guess it's mostly because of finishing up functional tests. I have a feeling the chapter could put a hard stop when we realised the first functional test passes. 

Apart from that, a potentially outdated part about possible typos in mock methods caught my eye. This definitely needs consideration.

### DIFF
--- a/chapter_21_mocking_2.asciidoc
+++ b/chapter_21_mocking_2.asciidoc
@@ -308,6 +308,9 @@ See <<car-factory-illustration>>.
 .Analysing how many tests are needed at different levels
 image::images/car-factory-illustration.png["An illustration of the car factory, with boxes for each step in the process (build engine, assemble, paint), and descriptions of testing each step separately vs testing them in combination."]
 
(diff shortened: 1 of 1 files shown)


chapter_21_mocking_2.asciidoc
```diff
@@ -443,6 +446,9 @@ because we've decided to use a particular, specific interface to implement our a
 which is something we might want to document and verify in our tests,
 and mocks are one way to enable that.
 
+// SEBASTIAN: I am missing one crucial sentence here - that this Django-provided abstraction IS STABLE, so it's safe to mock it.
+//      This is part of a public Django API, meaning it's not going anywhere soon or without breaking backwards-compatibility. That would of course be not welcomed by Django users :)
```
hjwp, line 450: nice point!  thank you.

PR #348, opened 2024-12-21
Csanád review of chapter 21

Xronophobe:
Although this was a longer chapter, I still found it easy to follow, and not overly verbose.


hjwp:
thanks - for future reference, it's probably easier to make any code suggestions outside of the code block.  as it is i probably have to revert them, then make a note to take a look later...

good suggestions tho!

hjwp:
hey it was no big deal in the end.  thanks for so thoughtfully separating out the commits buddy <3

PR #352, opened 2025-01-29
Update HTML tutorial link to MDN in pre-requisite-installations.asciidoc

DavidATN:
Removed link to webplatform.org (discontinued in 2015) and replaced with the new URL webplatform.org redirects to (MDN).

Might make it easy for newer developers to go straight to a free, reputable source


hjwp:
thanks David!  very much appreciated 😊

PR #356, opened 2025-02-16
Csanád review of chapter 23

Xronophobe:
The pace is, again, just right. Fast, not overly detailed but still easy to follow and answers pretty much all questions one may have. 

Later in the chapter, where we are deploying to the staging environment, I got stuck: `test_my_lists` failed by telling me it could not find `#id_logout`. Since the test ran fine just previously against the local docker, and the deployment is a `.tar` archive, and there may not be sqlite3 client installed by default, it may not be very straightforward to investigate if the reader believes there is a discrepancy between their local docker deployment and the staging.
This got me thinking: maybe we could add a few hints for what to do in case the reader believes there is a bug on staging: trying to reproduce it on the local docker environment and manually checking sqlite (e.g. by doing `SELECT * FROM account_token;` or `SELECT * FROM account_users;`...), add `logging.info()` or `logging.debug()` so that it shows up in the logs, etc). I think it would fit the scope of the chapter, as it's titled _Debugging And Testing Production Issues_.

### DIFF
--- a/chapter_23_debugging_prod.asciidoc
+++ b/chapter_23_debugging_prod.asciidoc
@@ -359,13 +359,21 @@ class FunctionalTest(StaticLiveServerTestCase):
 And you can confirm that the FT will fail if you don't set `EMAIL_PASSWORD` in Docker.
 
 Now let's see if we can get our FTs to pass against the server:
+// CSANAD: I think this is missing from here:
+// [subs="specialcharacters,macros"]
+// ----
+// $ pass:quotes[*TEST_SERVER=localhost:8888 python src/manage.py test functional_tests.test_login*]
+// [...]
+//
+// OK
+// ----
 
 
 === Setting Secret Environment Variables on the Server
 
 ((("environment variables")))
 ((("secret values")))
-Just as in <<chapter_11_server_prep>>,
+Just as in <<chapter_12_ansible>>,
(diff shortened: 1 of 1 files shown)


chapter_23_debugging_prod.asciidoc
```diff
@@ -328,8 +328,28 @@ Here's where we can put an early return in the FT:
     email = mail.outbox.pop()
 ----
 ====
-// CSANAD: TODO check if this would NOT return if we run the test against the
-// staging server, then?
+// CSANAD: This results in sending a real email to `edith@example.com` by
```
Xronophobe, line 331: @hjwp I received mails like this when I ran `test_login`. Can you check if you did too? I think you should have, because moving the early return higher up resolved it (clearly, since I'm no longer submitting the entered email address which caused the email to be sent) but I just want to cross-check this.

mailer_daemon

Xronophobe:
There is an error, which oddly says it's in either Chapter 18 or chapter 23 tests? I'm confused and I can't dig into it more deeply at the moment.

```
=================================== FAILURES ===================================
_____________ Chapter18Test.test_listings_and_commands_and_output ______________
tests/test_chapter_23_debugging_prod.py
```
Something about the `test_login_using_magic_link` and there is an error `500` - maybe whatever was set up for the CI to test this is down?

PR #360, opened 2025-02-20
Review of chapter 24

Enforcer:
This chapter is not the easiest one to read and follow.

I suggested adding some diagrams to visualize layers that we are crossing while doing outside-in TDD.

I would consider removing a part about `@property` or at least discourage readers from doing DB calls inside it.

### DIFF
--- a/chapter_24_outside_in.asciidoc
+++ b/chapter_24_outside_in.asciidoc
@@ -86,6 +86,10 @@ we start with the most outward-facing (presentation layer),
 through to the view functions (or "controllers"),
 and lastly the innermost layers, which in this case will be model code.
 
+// SEBASTIAN: I used grep to check when 'controller' appear in other places
+// and it seems its only in chapter 3 - once and a few times in this chapter.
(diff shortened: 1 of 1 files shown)


chapter_24_outside_in.asciidoc
```diff
@@ -1279,6 +1286,13 @@ even if it didn't have `@property`, but that's a particularity of Django, and
 doesn't apply to Python in general...
 *******************************************************************************
 
+// SEBASTIAN: While @property indeed is a helpful gimmick, I consider
+// @property doing DB operations or causing other side-effects an anti-pattern.
+// I wonder if readers of the book are also not already knowing that.
+// What I suggest is to consider whether to keep it in this chapter or not.
+// It seems to be a bit off. Might be as "quick hack" we're doing constantly to make
+// tests pass but I wouldn't settle on leaving it as it is.
```
hjwp, line 1294: good suggestion thanks, will revisit this.

PR #364, opened 2025-03-25
Review of debugging prod

seddonym:
Nice one. I particularly liked the discussion of when to choose _not_ to test something, in this case email.

I do think we should talk a little more about security here - readers should adopt a security mindset when doing things like this and make sure they think things through.

Note: I wasn't able to run the ansible commands as my Digital Ocean account has expired.


chapter_23_debugging_prod.asciidoc
```diff
@@ -464,6 +464,8 @@ FAILED (errors=1)
 
 It fails because our test utility function `create_pre_authenticated_session()`
 only acts on the local database.
+// DAVID: This feels like a logical leap. As a reader, I'm unsure whether I was supposed
+// to have known this, or if you're just saying 'trust me, I'll save you some time by telling you why'.
```
hjwp, line 468: thanks for this and all the comments david, really good to have this perspective.  will fix!

PR #366, opened 2025-03-26
Review of CI chapter

seddonym:


seddonym:
Another few of things maybe worth mentioning for additional context:

- Precommit hooks - maybe don't need to get into them here, but they do address some of the things that CI also addresses.
- Perhaps worth talking about how CI really comes into its own when working in teams, especially with open source. At the moment this is oriented around solo projects.
- Could mention linting too!

PR #371, opened 2025-04-16
Csanád's review of chapter 24

Xronophobe:
Very nice explanation on outside-in vs inside-out testing. I mentioned in a comment that maybe a simple sketching of what we want to see on the UI could be helpful before we get to write Functional Tests. This didn't occur to me in earlier chapters, which make sense - while initially every instruction about what the UI should look like was clear, as the UI gets more complex, it may be more difficult to clearly express it in just writing. Not crucial, but I think it would be nice.

 At the second part of the chapter, where we introduce the `owner` and then we move on with a failing test, I think we could have used a bit of a demonstration for a hybrid approach: by default we do outside-in, but in some cases it is perfectly fine to do some inside-out. Especially in the `test_views.py` where the test fails because the `owner` does not exist on the Model yet.

Seeing quite a few camelCase methods and functions made me think that maybe it would be nice to mention camelCase vs snake_case in chapter 02, where we first see a Python function using the camelCase scheme.

Overall, I really liked this chapter too.

### DIFF
--- a/chapter_24_outside_in.asciidoc
+++ b/chapter_24_outside_in.asciidoc
@@ -111,6 +111,10 @@ and lastly the innermost layers, which in this case will be model code.
 // and it seems its only in chapter 3 - once and a few times in this chapter.
 // Wouldn't it be easier for a reader to stick to Django's view?
 
+// CSANAD: just a thought: we already had some hand-written diagrams earlier.
+// What do you think about a super simple sketch of the UI before we extend the
+// functional test?
+
 We know our `create_pre_authenticated_session` code works now,
 so we can just fill out the actual body of our FT
 to look for a "My Lists" page:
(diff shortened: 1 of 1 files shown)


chapter_24_outside_in.asciidoc
```diff
@@ -111,6 +111,10 @@ and lastly the innermost layers, which in this case will be model code.
 // and it seems its only in chapter 3 - once and a few times in this chapter.
 // Wouldn't it be easier for a reader to stick to Django's view?
 
+// CSANAD: just a thought: we already had some hand-written diagrams earlier.
+// What do you think about a super simple sketch of the UI before we extend the
+// functional test?
```
hjwp, line 116: great idea!

PR #374, opened 2025-04-18
Csanád's review of chapter 26

Xronophobe:
This seemingly shorter chapter is actually quite a challenge (but does **not** give off the vibes of the meme known from college maths textbooks: "the proof is left for the reader as an exercise" :)  especially because the `book-example` repo is always there for cross-checking the work). The hints are very useful and basically showing what to do without revealing the implementation right away.
 I agree with the assumption of the reader should be able to test drive this feature on their own by now. And, if they do, it will be a very beneficial experience.


hjwp:
thanks buddy!  <3

PR #379, opened 2025-07-05
fix: update ai preface typo

rokbot:
Idk is this was intencional but that is a quick fix


hjwp:
thanks so much!  was just going to try and polish up this preface...

how are you enjoying the book so far?

PR #383, opened 2025-07-20
Csanád's review of chapter 27 and the AI Preface

CsanaadB:
I liked and I agree with the overview of the main lessons of the book (instead of main lessons, "key takeaways" would have been perfect, if it weren't for ChatGPT using it so much that now simply reading this one expression makes me feel like I'm reading LLM-generated word salad:) ).

 I was thinking if I should maybe update the `youtu.be/` links too, just to make them uniform `youtube.com/watch?v=` everywhere, but it wasn't really important after all.
 
 I don't have much to add to the AI Preface chapter: I really loved that you saw and drew attention to the importance of TDD when an LLM is generating the code. I don't think people usually think about that! 
Furthermore, I personally arrived to the same conclusions about the necessity of the human in the loop, with the code reviews and organizing the code base. 

<details>
<summary>
...Well, I had one additional thought, which I ended up not adding to the review.
</summary>

It's also about the human in the loop: 

For a while, I was on the opinion that no matter what, somebody has to take responsibility for the software. All software has some purpose, and its poor performance often has serious financial, reputational, economical, medical and criminal consequences. And taking responsibility isn't possible without understanding the code - so maybe "Junior" roles will be quite different, but "Senior" developers, and above, would be mainly unchanged in the foreseeable future.

Then, on a much more pessimistic tone, thinking about various scandals in IT security, safety and scandals from other various subjects, it's quite possible that a lack of understanding won't stop certain businesses from selling software nobody understands. I don't want to be the party pooper in your book, so I'll just leave it here, in a GitHub PR comment. :)

</details>


chapter_27_hot_lava.asciidoc
```diff
@@ -1,6 +1,7 @@
 [[chapter_27_hot_lava]]
 == Fast Tests, Slow Tests, and Hot Lava
 
+=======
```
hjwp:
```suggestion
```

PR #392, opened 2025-11-10
Fixed typos in chapter_21_mocking_2.asciidoc

Cactusson:


Cactusson:
I think it's fantastic. What I really like is how you describe, in great detail, the process of introducing an error and then fixing it (sometimes by googling the answer). This is something all other tutorials and books omit.

hjwp:
why thank you!  that is the kind of thing i'm proud of the book for.

hey it's been a while since i've asked this, but since the third edition is out, it could really do with a few reviews -- i don't suppose you'd be up for writing one? on amazon or goodreads or wherever you like to put them?

